PR #8, opened 2018-12-14
Adding update-channel.ps1

jcagme:
This will be the last step we run in the validation build and will add the build(s) which generated the Arcade SDK version we've just validated to the "Latest" channel

### DIFF
--- a/eng/validation/update-channel.ps1
+++ b/eng/validation/update-channel.ps1
@@ -1,0 +1,47 @@
+Param(
+  [string] $maestroEndpoint,
+  [string] $barToken,
+  [int] $channelId = 2             # channel 2 maps to  ".NET Tools - Latest" in Prod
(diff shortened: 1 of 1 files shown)


eng/validation/update-channel.ps1
```diff
@@ -0,0 +1,47 @@
+Param(
+  [string] $maestroEndpoint,
+  [string] $barToken,
+  [int] $channelId = 2             # channel 2 maps to  ".NET Tools - Latest" in Prod
+)
+
+. $PSScriptRoot\..\common\tools.ps1
```
riarenas: What are we getting from tools.ps1?

jcagme: It loads `$GlobalJson` with the contents to `global.json`

PR #10, opened 2018-12-17
Refer execution of helix.yml to jobs.yml

jcagme:
Tes build: https://dnceng.visualstudio.com/internal/_build/results?buildId=59803

### DIFF
--- a/eng/validation/templates/testing/helix.yml
+++ b/eng/validation/templates/testing/helix.yml
@@ -1,32 +1,27 @@
 jobs:
-  - job:
-    displayName: Validate Helix
-    variables:
-      - group: DotNet-HelixApi-Access
-      - name: _BuildConfig
-        value: Release
-    steps:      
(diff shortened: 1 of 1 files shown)


eng/validation/templates/testing/helix.yml
```diff
@@ -1,32 +1,27 @@
...
-      condition: always()
\ No newline at end of file
+  - template: /eng/common/templates/jobs/jobs.yml
+    parameters:
+      enablePublishBuildArtifacts: true
+      jobs:
+        - job:
+          displayName: Validate Helix
+          variables:
+            - group: DotNet-HelixApi-Access
```
riarenas: If we're using jobs.yml. We can use the variable:value syntax right?

jcagme: Hmm that's what I'm using bellow (?)

riarenas: You are indeed. Not sure what I was looking at 😮

PR #12, opened 2018-12-18
add final build job that updates the channel, and only runs if every other job succeeds

riarenas:
fixes dotnet/arcade#1612


riarenas:
Example build: https://dnceng.visualstudio.com/internal/_build/results?buildId=60361

riarenas:
Build after changes: https://dnceng.visualstudio.com/internal/_build/results?buildId=60543

PR #22, opened 2018-12-19
Improve the update-channel.ps1 script

jcagme:


eng/validation/update-channel.ps1
```diff
@@ -15,34 +15,45 @@ function Get-Headers([string]$accept, [string]$barToken) {
...
+$assets = Invoke-WebRequest -Uri $getAssetsApiEndpoint -Headers $headers | ConvertFrom-Json
+
+if (!$assets) {
     Write-Host "Asset '$arcadeSdkPackageName' with version $arcadeSdkVersion was not found"
     exit 1
-}
\ No newline at end of file
+}
+
+if ($assets.Count > 1) {
```
mmitche: I think this should be -ne 1.  Although I'm not positive, it's possible the assets could just be an empty json array. In that case, we'd pass through the above !$assets and this statement.

eng/validation/update-channel.ps1
```diff
@@ -15,34 +15,45 @@ function Get-Headers([string]$accept, [string]$barToken) {
...
+$matchedChannel = $($channels | Where-Object -Property "name" -Value "${targetChannelName}" -EQ | Select-Object -Property id)
+
+if (!$matchedChannel) {
+    Write-Host "Channel with name '$targetChannelName' was not found..."
+    exit 1    
+}
+
+$channelId = $matchedChannel.id
+
+try {
```
mmitche: Why the try/catch here but not around the other calls?

jcagme: When a build already exists in a channel we are returning 500 which for Invoke-WebRequest is considered an exception. Given the refactor we could just let the call fail since we are not in a loop anymore

eng/validation/update-channel.ps1
jcagme: Just validated and `!$assets` works fine with empty arrays

eng/validation/update-channel.ps1
mmitche: Okay.  I have an issue I just opened re: that issue, will fix.

azure-pipelines.yml
```diff
@@ -120,3 +120,4 @@ jobs:
       - powershell: eng/validation/update-channel.ps1
           -maestroEndpoint https://maestro-prod.westus2.cloudapp.azure.com
           -barToken $(MaestroAccessToken)
+        displayName: Move build to 'Latest' channel
```
riarenas: There's already a displayName for this one in LN 108

eng/validation/update-channel.ps1
```diff
@@ -15,34 +15,45 @@ function Get-Headers([string]$accept, [string]$barToken) {
...
+    $buildId = $assets[0].'buildId'
+
+    $postBuildIntoChannelApiEndpoint = "$maestroEndpoint/api/channels/$channelId/builds/$buildId/?name=$arcadeSdkPackageName&version=$arcadeSdkVersion&api-version=2018-07-16"
+    $headers = Get-Headers 'application/json' $barToken
+        
+    Write-Host "POSTing to $postBuildIntoChannelApiEndpoint..."
+    $postResponse = Invoke-WebRequest -Uri $postBuildIntoChannelApiEndpoint -Headers $headers -Method Post
```
riarenas: As discussed offline. One further improvement is to check if the build is already in the Latest channel. If it is, we shouldn't try to post it, and it shouldn't be considered an error.

azure-pipelines.yml
jcagme: If you expand the "Push Builds to 'Latest' channel" in https://dnceng.visualstudio.com/internal/_build/results?buildId=61056 you'll see this step is shown as "Powershell" only. In this test build you can see it displays "Move build to 'Latest' channel" https://dnceng.visualstudio.com/internal/_build/results?buildId=61149

eng/validation/update-channel.ps1
jcagme: https://dnceng.visualstudio.com/internal/_build/results?buildId=61149 implements this

azure-pipelines.yml
riarenas: I missed that this is for the step. doh.

eng/validation/update-channel.ps1
riarenas: Is that code in this branch? I don't see those messages in the PR.

PR #26, opened 2018-12-19
Use open queues during PR, non-open during official build.

riarenas:
We now do the send_to_helix job in both PR and official builds.

PR job uses anonymous mode, Official build uses authenticated mode.

fixes dotnet/arcade#1641


riarenas:
Waiting for https://dnceng.visualstudio.com/internal/_build/results?buildId=61384 to complete to ensure this fixes the Helix leg.

PR #43, opened 2019-01-02
Add nuget version to Versions.props (TEST)

riarenas:
This seems to be the cause of nuget incorrectly restoring the wrong arcade SDK version.

Internal builds using this change haven't failed so far, but I want to test this using the external flow as well.


riarenas:
At worst this won't do anything, and at worst it'll solve the problem. I'm going to merge and keep an eye out before concluding this is fixed.

PR #44, opened 2019-01-03
Remove sign and testing templates and refactor yaml

riarenas:
We're not getting much out of templates: We don't reuse them at all, and it's just leading to repeated and unnecessary verbosity across multiple files.

This change removes the template approach, and uses the pattern we're using in other repos: Having all jobs specified in the base yaml.

If we add a leg that actually benefits from a separate template than the ones Arcade provides out of the box, we can reintroduce them when the time comes.

Also removes some lines that we accidentally copied over from Arcade's definition, and weren't used in ours.


riarenas:
Clarification: We're not getting much from templates in the validation repo. We were basically using them to encapsulate jobs, while their main purpose is to foster reuse.

azure-pipelines.yml
```diff
@@ -45,7 +39,7 @@ jobs:
             /p:DotNetSymbolServerTokenMsdl=$(microsoft-symbol-server-pat)
             /p:DotNetSymbolServerTokenSymWeb=$(symweb-symbol-server-pat)
             /p:OfficialBuildId=$(BUILD.BUILDNUMBER)
-
```
chcosta: The DotNetSymbolServerTokenMsdl and DotNetSymbolServerTokenSymWeb variables are not needed.  I'm removing them from Arcade in https://github.com/dotnet/arcade/pull/1595/files#diff-fec826feae04e51c0d94076385408bdc

azure-pipelines.yml
```diff
@@ -100,24 +94,85 @@ jobs:
           --configuration $(_BuildConfig)
           --prepareMachine
         displayName: Unix Build / Publish
-- template: /eng/validation/templates/testing/helix.yml
```
chcosta: "_PublishType"  parameter isn't used anywhere (same reference where I'm removing this from Arcade as the above comment).

PR #74, opened 2019-01-14
add default for helix access token

riarenas:
Sending to Helix is broken without setting this parameter first.


mmitche:
@chcosta Does this look good?  Arcade flow is broken without this fix.

riarenas:
Ugh. Looks like my email about this never got sent. Thanks for looking at it @mmitche and @chcosta.

PR #101, opened 2019-01-24
Disable push to latest channel

riarenas:
Temporary workaround. This should allow us to keep validating newer Arcade builds, but we won't push them to the channel where other repos would consume from.

@markwilkie @jcagme @mmitche

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -151,6 +151,7 @@ jobs:
               /p:OfficialBuildId=$(BUILD.BUILDNUMBER)
             displayName: Sign packages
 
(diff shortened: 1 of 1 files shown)


riarenas:
This will also guard us from someone merging a PR to Arcade that would unblock the arcade validation build and then start flowing versions we don't want and is less fragile than just hoping a fix for the current blocker doesn't make it in.

jcagme:
We should think of a way of making this configurable instead of comment/uncomment code in the future :) Probably moving build to a "Limbo" channel by just passing -targetChannelName Limbo

jcagme:
Or that, just going "up" instead of going "down: :)

PR #104, opened 2019-01-29
re-enable push to latest channel

riarenas:
Re-enabling now that P2 shipped.

@markwilkie, @jcagme


riarenas:
No azdo gate is running. Possibly related to the current widespread issues? Will hold on on this one until we can validate CI.

PR #348, opened 2019-04-30
update sdk to avoid nuget bug, and move to prod BYOC provider

riarenas:
* Move to use prod pool provider
* Use newer version of SDK to try and prevent NU3004 error


riarenas:
Will keep an eye out to see if this helps with the NU3004 errors we've been seeing.

PR #487, opened 2019-06-06
Fixes for arcade validation build

riarenas:
Fixes https://github.com/dotnet/arcade/issues/2971

* Stop using the int pool provider
* The Validate_Signing and SDL jobs are now part of the jobs collection from the jobs.yml templates, so that they are required to succeed before the publish to BAR job.
* The push to 'latest' channel job now only depends only on the publish to BAR job.


riarenas:
Validation build for this: https://dev.azure.com/dnceng/internal/_build/results?buildId=214946

PR #494, opened 2019-06-06
only do the push build to latest channel job if running from master b…

riarenas:
…ranch

fixes https://github.com/dotnet/arcade/issues/2993

* Also adds back the build step in the SDL job


riarenas:
sample build that doesn't run the job: https://dev.azure.com/dnceng/internal/_build/results?buildId=215460

PR #610, opened 2019-07-09
Copy sources from Versions.props to NuGet.config

jcagme:
**TL;DR**

We need that all restore sources are located in `NuGet.config` since internal feeds can only be restored from there. More details including next steps [here](https://github.com/dotnet/arcade/blob/master/Documentation/RestoreSourcesUpdateStatus.md)

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -9,8 +9,8 @@
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
     <add key="arcade-validation" value="https://dotnetfeed.blob.core.windows.net/arcade-validation/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
-    <add key="arcade-validation" value="https://dotnet.myget.org/F/symreader-converter/api/v3/index.json" />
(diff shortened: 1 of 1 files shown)


NuGet.config
```diff
@@ -9,6 +9,8 @@
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
     <add key="arcade-validation" value="https://dotnetfeed.blob.core.windows.net/arcade-validation/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="arcade-validation" value="https://dotnet.myget.org/F/symreader-converter/api/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
riarenas: nuget.org is already there, and arcade-validation is a key already

riarenas: It also looks like we're missing  https://dotnet.myget.org/F/symreader/api/v3/index.json

jcagme: Oh yeah, I was in the middle of the update when you dropped by and I almost got a heart attack :)

riarenas: Always Ricardo's fault 😭

PR #663, opened 2019-07-26
Add multi-native tool install testing

jonfortescue:
Will test https://github.com/dotnet/arcade/pull/3400


riarenas:
`Failed to download 'https://netcorenativeassets.blob.core.windows.net/resource-packages/external/linux/cmake/cmake-3.11.1jdk:10.0.2-Linux-x86_64.tar.gz`

😢

jonfortescue:
AAAAAAAAA

jonfortescue:
@riarenas looking at the build logs more carefully, the windows build *should* have failed but didn't (it failed to install the jdk). Any ideas?

riarenas:
install-tools.ps1 doesn't set any ErrorActionPreference. Maybe that's what's causing it to not fail the build?

eng/common/init-tools-native.ps1
```diff
@@ -44,6 +44,9 @@ Param (
   [string] $GlobalJsonFile
 )
 
+Set-StrictMode -Version 2.0
```
riarenas: My only concern with this (when it goes to Arcade) is that this would change current behavior, and although it's probably the right thing to do,  it wouldn't surprise me at all if someone already depends on the fact that failures don't cause build breaks.

jonfortescue: :\ yeah you're right this would be a breaking change. I'll revert this for now and file an issue.

jonfortescue: Filed https://github.com/dotnet/arcade/issues/3502 to deal with this.

jonfortescue: I misread and did not touch `install-tools.ps1`. Idk what's going on here but init-tools-native.ps1 [already has](https://github.com/dotnet/arcade/blob/master/eng/common/init-tools-native.ps1#L51-L52) these in it.

PR #758, opened 2019-08-16
Add proper CI trigger to azure-pipelines.yml

alexperovich:
Without this the CI build runs for every pushed branch ever.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,6 +8,12 @@ variables:
   - name: _DotNetValidationArtifactsCategory
     value: .NETCoreValidation
 
+trigger:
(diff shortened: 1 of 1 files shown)


JohnTortugo:
My initial thought was that this could somehow affect the promotion of Arcade builds to Latest channel, but after some thought and a quick investigation in `update-channel.ps1` I think it should be OK. What do you think @riarenas ?

alexperovich:
The lack of proper filter is causing the darc-master-<guid> branches to get builds ran for them. They are all useless builds. Essentially a PR gets built twice.

riarenas:
Yep this won't interfere with anything.

PR #862, opened 2019-09-16
[release/3.x] Update target channel

mmitche:


azure-pipelines.yml
```diff
@@ -13,6 +13,7 @@ trigger:
   branches:
     include:
       - master
```
JohnTortugo: Do we need to remove `master` here?

mmitche: nope

PR #866, opened 2019-09-19
move copy-to-latest-channel to its own stage

riarenas:
Fixes the gap outlined in: https://github.com/dotnet/arcade/issues/3982

Test build incoming once the fix for the arcade templates makes it to the latest channel but want to get early feedback on the approach.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -210,8 +210,6 @@ stages:
         displayName: Push Builds to 'Latest' channel
         variables:
           - group: Publish-Build-Assets
-        dependsOn:
-          - Asset_Registry_Publish
         steps:
           - checkout: self
             clean: true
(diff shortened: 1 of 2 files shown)


riarenas:
We will also need this change in release/3.x, so once I have validated the fix I will create an ask mode template for it.

riarenas:
Test build https://dev.azure.com/dnceng/internal/_build/results?buildId=360262&view=results looks how I expect it to

azure-pipelines.yml
```diff
@@ -210,3 +193,28 @@ stages:
         -TsaRepositoryName "Arcade-Validation"
         -TsaCodebaseName "Arcade-Validation"
         -TsaPublish $True'
+
+  - stage: Push_to_latest_channel
```
chcosta: maybe specify a display name?

riarenas: Yep, that was in my to do list. Thanks for the reminder

PR #886, opened 2019-09-24
Re-add removed file

ViktorHofer:
Was already brought back by https://github.com/dotnet/arcade-validation/commit/f9f843041d144c74f79668e2fef1e762e983509f but deleted in the subsequent commit. cc @riarenas


chcosta:
@alexperovich @riarenas , can you PTAL at why this file keeps getting deleted.  Seems like something is up with darc.

PR #905, opened 2019-10-02
Add the dotnet-tools feed for restores

riarenas:
Test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=374380

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -7,6 +7,7 @@
   <packageSources>
     <clear />
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="arcade-validation" value="https://dotnetfeed.blob.core.windows.net/arcade-validation/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="symreader-converter" value="https://dotnet.myget.org/F/symreader-converter/api/v3/index.json" />


ViktorHofer:
Is that feed a real nuget feed?

riarenas:
It is! Builds that get published to the "dotnet tools - latest" channel publish to that feed.

Mmitche has been adding the feed to release branches for most (all?) product repos.

mmitche:
> Is that feed a real nuget feed?

It's real! You can even browse it.

ViktorHofer:
Unrelated, but would that feed be suitable for VSTest to push to?

mmitche:
no. vstest should push to dotnet3, dotnet3-transport, dotnet5, dotnet5-transport, dotnet3.1, dotnet3.1-transport, ideally through the arcade publishing process.

the -transport packages are for non-shipping packages that will never have stable versions (3.0.0-perview-1234)

ViktorHofer:
VSTest currently can't use the arcade publishing process as it's not fully onboarded yet and they don't have resources to do so. 

> no. vstest should push to dotnet3, dotnet3-transport, dotnet5, dotnet5-transport, dotnet3.1, dotnet3.1-transport, ideally through the arcade publishing process.

As they don't depend on a specific .NET Core version and because they don't compile against the always latest one, wouldn't a non versionless feed make more sense? at the moment I recommend to push to the dotnet-core azure blob feed.

ViktorHofer:
Updated typos...

mmitche:
> As VSTest doesn't depend on a specific .NET Core version and because they don't compile against the always latest one, wouldn't a versionless feed make more sense? I recommended to push to the dotnet-core azure blob feed for now.

We can create one for them if they'd like.  dotnet-tools is really only for internal tools though.

ViktorHofer:
That would work. I'm surprised that we don't have a general purpose feed for partner teams yet (which don't use Arcade).

PR #934, opened 2019-10-21
Remove blob feed from nuget.config

mmitche:
This caused a problem when recently arcade stopped publishing to dotnet-tools because of an issue with the order a change should have been made. https://github.com/dotnet/arcade/pull/4117

Because arcade-validation was in this nuget.config, this problem was not caught immediately, and only seen when arcade was ingested into itself.


NuGet.config
```diff
@@ -8,7 +8,6 @@
     <clear />
     <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
```
riarenas: Should we also remove this one? I don't think there's anything we still expect to only be in the blob feed.

PR #939, opened 2019-10-22
Update AreaPath for SDL TSA Onboarding

sunandabalu:
Area Path in DevDiv changed for NET Core, updating the settings to change the same. 
Test Build - https://dev.azure.com/dnceng/internal/_build/results?buildId=397080&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76&j=7d9eef18-6720-5c1f-4d30-89d7b76728e9

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,6 +7,8 @@ variables:
     value: .NETCore
   - name: _DotNetValidationArtifactsCategory
     value: .NETCoreValidation
+  - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+    - group: SDL_Settings
 
 trigger:
   batch: true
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -188,7 +188,7 @@ stages:
         -TsaProjectName "DEVDIV"
         -TsaNotificationEmail "dnceng@microsoft.com"
         -TsaCodebaseAdmin "REDMOND\dn-bot"
-        -TsaBugAreaPath "DevDiv\NET\NET Core "
+        -TsaBugAreaPath "DevDiv\NET Core "
```
riarenas: I wonder if it makes sense to make this a default value that Arcade populates. That way we could control similar errors that happen underneath us by fixing it in Arcade and flowing to all repos. (Not saying we should do that as part of this change, but maybe open an issue to evaluate the feasibility?)

sunandabalu: Yep opening an issue to evaluate this is a good idea.

PR #1005, opened 2019-11-08
*NO MERGE* Testing staging build pools

jcagme:


riarenas:
@jcagme Do you still want to keep this open? We now have a daily run of this repo that builds using the staging pool provider so not sure if that takes care of what you're testing here?

PR #1016, opened 2019-11-11
Use PoolProvider, supplied as a queue-time parameter

adiaaida:
The parameter has already been added to the public and internal pipelines.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ variables:
   - ${{ if and(eq(variables.PoolProvider, ''), eq(variables['System.TeamProject'], 'public')) }}:
     - name: PoolProvider
       value: NetCorePublic-Pool
-  - ${{ if and(eq(variables.PoolProvider, ''), ne(variables['System.TeamProject'], 'public')) }}:
+  - ${{ if and(ne(variables.PoolProvider, ''), ne(variables['System.TeamProject'], 'public')) }}:
     - name: PoolProvider
       value: NetCoreInternal-Pool
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,6 +9,12 @@ variables:
     value: .NETCoreValidation
   - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
     - group: SDL_Settings
+  - ${{ if and(eq(variables.PoolProvider, ''), eq(variables['System.TeamProject'], 'public')) }}:
+    - name: PoolProvider
+    - value: NetCorePublic-Pool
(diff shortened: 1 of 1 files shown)


adiaaida:
I have done it this way, but I am worried that it isn't quite right because the logic for the first if is an or of public and pull-request. So if there is a PR to the internal side, it will now get the internal pool provider, which is wrong. I'm wondering if I should actually default the PoolProvider parameter to empty, and in the code here, do:

```
if and(PoolProvider != '', or(public, pr):
poolprovider = public pool
if and(poolprovider != '', and(internal, not pr):
poolprovider = internal pool
```
and then when we add the scheduled job/separate pipeline(?) set the pool provider to staging.

alexperovich:
Why is an internal PR getting the internal pool provider wrong? An internal PR should always get the internal pool provider because it could be building internal bits that we can't make public.

---
In reply to: [552638539](https://github.com/dotnet/arcade-validation/pull/1016#issuecomment-552638539) [](ancestors = 552638539)

adiaaida:
Then the current code is wrong?

`${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}: # public pool`

`${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}: # internal pool`

alexperovich:
Yea... I think it is. It actually won't work as it stands. The internal project doesn't have the public pool available. We haven't needed to do any internal PRs though.

---
In reply to: [552644714](https://github.com/dotnet/arcade-validation/pull/1016#issuecomment-552644714) [](ancestors = 552644714)

sunandabalu:
I see that currently the PoolProvider value is set to NetCorePublic-Pool for both arcade-validation-official(internal) and arcade-validation-ci(public), should they not be different? The change itself LGTM.

adiaaida:
Thanks for the catch, @sunandabalu. I'd changed it but forgot to save.

MattGal:
> Thanks for the catch, @sunandabalu. I'd changed it but forgot to save.

Is this reflected in the current version?  Should definitely not merge in current state. I'd favor two variables for the two different providers.

adiaaida:
@MattGal I can add a default. Also, yes, I have set this variable for all of the arcade-validation pipelines.

adiaaida:
@MattGal I added defaults in case the variable hasn't been set. Let me know if this takes care of that for you and I'll check in.

MattGal:
> @MattGal I added defaults in case the variable hasn't been set. Let me know if this takes care of that for you and I'll check in.

Yup I already approved, as long as you're on point to make sure that the builds keep working let's merge!

PR #1320, opened 2020-03-04
Bug: there isn't a BuildPool.Ubuntu.1604.Amd64.Arcade queue

MattGal:


MattGal:
This line only impacts internal builds and is needed so I am merging to unbreak builds.

PR #1323, opened 2020-03-04
Trying Hosted Ubuntu 1604 for builds

MattGal:


MattGal:
(internal build where this worked: https://dev.azure.com/dnceng/internal/_build/results?buildId=545720&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76 )

MattGal:
> We still get coverage for both int and prod pool providers via the windows builds, so I see no reason not to make this change if it alleviates some burden on the ubuntu buildpools.

Good point, I'm going to make it only used hosted for internal builds.

PR #1360, opened 2020-03-20
Removing automatic Arcade flow; including Roslyn in Arcade Validation

missymessa:


azure-pipelines.yml
```diff
@@ -204,38 +204,34 @@ stages:
...
-      - General_Testing_Publish
-      - NETCore_Tooling_Dev_Publishing
-      - NETCore_Tooling_Release_Publishing
-      - NET_Internal_Tooling_Publishing
-      - NETCore_Experimental_Publishing
-      - Net_Eng_Services_Int_Publish
-      - Net_Eng_Services_Prod_Publish
-    jobs:
+# Arcade validiation with additional repos
+- stage: Additional_Arcade_Validation
```
riarenas: This probably needs a more descriptive stage and displaynames. having a stage called addition validation in a build that already has a validate stage in a repo called arcade-validation. It's like the word has lost all meaning.

Something like validate_consumer_repos?

missymessa: image

azure-pipelines.yml
```diff
@@ -204,38 +204,34 @@ stages:
...
+  displayName: Additional Arcade Validation
+  jobs:
     - template: /eng/common/templates/job/job.yml
       parameters:
-        name: Push_to_latest_channel
```
riarenas: I think this will save you some pain in the future if you declare a job matrix here so that we can just keep adding additional repos. I won't block on it this time, but next

missymessa: Agreed. I was going to configure that when I did the runtime repo.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+
+$arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
+$arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$arcadeSdkPackageName
+
+# Get a temporary directory for a test root. Use the agent work folder if running under azdo, use the temp path if not.
+$testRootBase = if ($env:AGENT_WORKFOLDER) { $env:AGENT_WORKFOLDER } else { $([System.IO.Path]::GetTempPath()) }
+$testRoot = Join-Path -Path $testRootBase -ChildPath $([System.IO.Path]::GetRandomFileName())
+New-Item -Path $testRoot -ItemType Directory | Out-Null
+
+$days = 3
```
riarenas: Should we make this an argument to the script? Can still have this default value.

azure-pipelines.yml
```diff
@@ -204,38 +204,34 @@ stages:
...
-      - Net_Eng_Validation_Publish
-      - General_Testing_Publish
-      - NETCore_Tooling_Dev_Publishing
-      - NETCore_Tooling_Release_Publishing
-      - NET_Internal_Tooling_Publishing
-      - NETCore_Experimental_Publishing
-      - Net_Eng_Services_Int_Publish
-      - Net_Eng_Services_Prod_Publish
-    jobs:
+# Arcade validiation with additional repos
```
mmitche: Nit: validation.

azure-pipelines.yml
```diff
@@ -204,38 +204,34 @@ stages:
...
+              -githubPAT $(BotAccount-dotnet-bot-repo-PAT)
+              -githubOrg "dotnet"
+              -targetRepoName "roslyn"
               -barToken $(MaestroAccessToken)
-              -azdoToken $(dn-bot-dnceng-build-rw-code-rw)
-              -githubToken $(BotAccount-dotnet-bot-repo-PAT)
-            displayName: Move build to 'Latest' channel 
+              -buildParameters '{"SkipApplyOptimizationData":"true"}'
+  
+# End Arcade validation with runtime and roslyn
```
riarenas: [nit] with additional repos to match the other comment?

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    ## No? Then get the last known good build regardless of age
+    if($count -eq 0)
+    {
+        $contentArray = @()
+
+        foreach($reason in $buildReasonsList)
+        {
+            $uri = "https://dev.azure.com/${azdoOrg}/${azdoProject}/_apis/build/builds?api-version=5.1&resultsFilter=succeeded&definitions=${buildDefinitionId}&reasonFilter=${reason}&buildQueryOrder=finishTimeAscending&`$top=1"
```
riarenas: This can probably be extracted into a function as it's repeated below just with slightly different params.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+    ## Have there been any builds in the last $days days? 
+    $base64AuthInfo = [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(":${azdoToken}"))
+    $headers = @{"Authorization"="Basic $base64AuthInfo"}
+    $count = 0
+
+    foreach($reason in $buildReasonsList)
+    {
+        $uri = "https://dev.azure.com/${azdoOrg}/${azdoProject}/_apis/build/builds?api-version=5.1&statusFilter=completed&definitions=${buildDefinitionId}&reasonFilter=${reason}&minTime=${minTime}&`$top=1"
+        $reponse = (Invoke-WebRequest -Uri $uri -Headers $headers -Method Get) | ConvertFrom-Json
+        $count += $reponse.count
```
riarenas: Is there a reason not to just get the build already? (as in LN95)

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+$buildReasonsList = @("batchedCI", "individualCI")
+
+function Get-LastKnownGoodBuildSha(
+    [string] $azdoOrg, 
+    [string] $azdoProject,
+    [int] $buildDefinitionId,
+    [string] $azdoToken)
+{
+
+    ## Have there been any builds in the last $days days?
```
riarenas: Should be able to extract this into a Set-Headers function or something. This is duplicated in the Invoke function below.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+
+function Get-Github-RepoAuthUri($repoName)
+{
+    "https://${githubUser}:${githubPAT}@github.com/${githubOrg}/${repoName}"
+}
+
+function GitHub-Clone($repoName) 
+{
+    $authUri = Get-Github-RepoAuthUri $repoName
+    & git clone $authUri $(Get-Repo-Location $repoName)
```
riarenas: Should this also use Git-Command?

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+$githubUri = Get-Github-RepoAuthUri $targetRepoName
+
+$darcBranchName = "refs/heads/" + $targetBranch
+$darcRepoName = "https://github.com/${githubOrg}/${targetRepoName}"
+
+# check to see if branch exists and clean it up if it does
+$branchExists = Git-Command $targetRepoName ls-remote --heads $githubUri refs/heads/$targetBranch
+if($null -ne $branchExists)
+{
+    & darc delete-default-channel --channel "General Testing" --branch $darcBranchName --repo $darcRepoName --github-pat $githubPAT --password $barToken
```
riarenas: We probably want to make sure that this doesn't fail the script. There's a (very very short) race where someone could delete the branch right before we check.

Same with default channel, someone could've deleted it.

If we fail to delete them we should inform, but not fail.

missymessa: Are there specific errors to look for from darc, or would wrapping this in a try/catch be sufficient?

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+## Make the changes to that branch to update Arcade - use darc
+Set-Location $(Get-Repo-Location $targetRepoName)
+& darc update-dependencies --channel ".NET Eng - Validation" --source-repo "arcade" --github-pat $githubPAT --password $barToken
+
+## Push branch to github
+Git-Command $targetRepoName commit -am "Arcade Validation test branch - version ${arcadeSdkVersion}"
+Git-Command $targetRepoName push origin HEAD
+
+## Push branch to AzDO org/project with the official pipeline to build the repo
+## Don't need to do this for Roslyn, but we'll need to do it for Runtime.
```
riarenas: Not only that, but we should probably not do it in GitHub at all for anything not roslyn.

azure-pipelines.yml
missymessa: Renamed to "Validate Arcade with Consumer Repositories"

azure-pipelines.yml
missymessa: Got rid of this line.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+}
+
+## Output summary of references for investigations
+Write-Host "Arcade Version: ${arcadeSdkVersion}"
+Write-Host "Repository Cloned: ${githubOrg}/${targetRepoName}"
+Write-Host "Branch name in repository: ${targetBranch}"
+Write-Host "Last Known Good build SHA: ${sha}"
+
+## If build fails, then exit
+if("failed" -eq (Get-BuildResult -azdoOrg $azdoOrg -azdoProject $azdoProject -buildId $buildId -azdoToken $azdoToken))
```
riarenas: We should also check for cancelled here. Someone could fat finger and cancel the build, and we probably don't want to take that as success.

eng/validation/build-arcadewithrepo.ps1
riarenas: Try/catch is basically what I was thinking

eng/validation/build-arcadewithrepo.ps1
missymessa: Line 95 is specifically looking for successful builds in the last three days. This first check is to see if there are any completed builds (regardless of success or fail or whatever is in between). If you think the API calls are too expensive, I can get all the completed builds on 95 and then filter through them to see which are successes or non-successes.

riarenas: Right, that nuance was still confusing me lol. The API call is not expensive so it's fine to make it.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+{
+    $authUri = Get-Github-RepoAuthUri $repoName
+    & git clone $authUri $(Get-Repo-Location $repoName)
+    Push-Location -Path $(Get-Repo-Location $repoName)
+    & git config user.email "${githubUser}@test.com"
+    & git config user.name $githubUser
+    Pop-Location
+}
+
+function Get-Repo-Location($repoName)
```
JohnTortugo: NIT: perhaps this would be better if inlined.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+## Don't need to do this for Roslyn, but we'll need to do it for Runtime. 
+
+## Add default channel from that github repo and branch to "General Testing"
+& darc add-default-channel --channel "General Testing" --branch $darcBranchName --repo $darcRepoName --github-pat $githubPAT --password $barToken
+
+## Run an official build of the branch using the official pipeline
+$buildId = Invoke-AzDOBuild -azdoOrg $azdoOrg -azdoProject $azdoProject -buildDefinitionId $buildDefinitionId -azdoToken $azdoToken -branchName $targetBranch -buildParameters $buildParameters
+
+## Check build for completion every 5 minutes. 
+while("completed" -ne (Get-BuildStatus -azdoOrg $azdoOrg -azdoProject $azdoProject -buildId $buildId -azdoToken $azdoToken))
```
JohnTortugo: Would be nice to print the link to the build at least once so that we can easily click + inspect if needed.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
...
+Write-Host "Last Known Good build SHA: ${sha}"
+
+## If build fails, then exit
+if("failed" -eq (Get-BuildResult -azdoOrg $azdoOrg -azdoProject $azdoProject -buildId $buildId -azdoToken $azdoToken))
+{
+    Write-Error "Build failed"
+    exit
+}
+
+## Clean up branch if successful
```
JohnTortugo: Don't we need to run these when things fail? Perhaps would be useful to create a "finally {...} " block to make sure a clean up is always done.

missymessa: I thought it might be useful to keep the branch around for investigation purposes if there is a failure. @mmitche do you have an opinion on whether keeping these around would be useful or not? (I have an issue to write a clean up script when we're finished with investigations).

eng/validation/build-arcadewithrepo.ps1
missymessa: Oops, nope, it's not going to work. Git-Command takes in the repoName and calls Push-Location on it, but since it doesn't exist before the clone, it errors out :(

eng/validation/build-arcadewithrepo.ps1
missymessa: Done!

eng/validation/build-arcadewithrepo.ps1
missymessa: Good idea :)  I added it.

eng/validation/build-arcadewithrepo.ps1
missymessa: Get-Headers all the things :)

eng/validation/build-arcadewithrepo.ps1
missymessa: Now checking for failed or cancelled builds

eng/validation/build-arcadewithrepo.ps1
missymessa: Done.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -0,0 +1,265 @@
+Param(
+  [Parameter(Mandatory=$true)]
+  [string] $azdoOrg, 
+
```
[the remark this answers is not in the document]
missymessa: Done.

eng/validation/build-arcadewithrepo.ps1
missymessa: Did that here and also on the clean up at the end of the script just in case.

eng/validation/build-arcadewithrepo.ps1
missymessa: I extracted most of the uri into it's own function, and almost started to do the same with the query string parameters, but the function was going to have to take in so many parameters that it would've defeated the purpose of creating a method to return a string. Hope that's good enough. :/

riarenas: What I meant was that you could extract the whole call to the API to a function. 

https://github.com/dotnet/arcade-validation/pull/1360/files#diff-0c4c1d2780faad03a492cdb1cab92517R52-R60 and https://github.com/dotnet/arcade-validation/pull/1360/files#diff-0c4c1d2780faad03a492cdb1cab92517R71-R79 are duplicate. You just change the params to the build API and what you do with the resulting array.

riarenas: That would also allow us to add retries, etc in a single place if we see flakiness.

eng/validation/build-arcadewithrepo.ps1
missymessa: Updated the task list in the other issues for the other repos to not include pushing to GitHub.

azure-pipelines.yml
missymessa: Add this as a comment in core-eng#9283

eng/validation/build-arcadewithrepo.ps1
missymessa: I'm going to resolve this so I can merge this PR. If we decide to always clean up because we don't need the branches for investigation, I'm happy to change it :)

ViktorHofer:
@missymessa Why was this done? Arcade is stale since Saturday:

image

markwilkie:
In order to help stabilize net 5 builds, we've gone to a manual arcade flow.  In addition, we've updated the policy so we don't "pile on" changes when the repos are broken.  See https://github.com/dotnet/arcade/blob/master/Documentation/Validation/Overview.md

As noted, fixed needed to unblock will of course be flowed - which is why I reached out to you separately.

cc/ @jaredpar

eng/validation/build-arcadewithrepo.ps1
missymessa: Sorry Ricardo, I somehow completely missed these comments. I can combine those two, but there's a third API call in the same method with very similar (but different) parameters: https://github.com/dotnet/arcade-validation/pull/1360/files#diff-0c4c1d2780faad03a492cdb1cab92517R42

Are you okay with that one being left out of a method that wraps those two API calls into a method? I was trying to figure out how to get all three of them into a method and since they were taking different parameters, the method call was going to be as long as the query string parameter (probably longer, because this is Powershell, of course) :)

riarenas: Yeah, I'm fine with that.

PR #1367, opened 2020-03-26
Adding Runtime to Arcade Validation process

missymessa:
(I also updated Arcade in this branch)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -209,26 +209,42 @@ stages:
     jobs:
       - template: /eng/common/templates/job/job.yml
         parameters:
-          name: Validate_Arcade_with_Roslyn
-          displayName: Validate Arcade with Roslyn
-          timeoutInMinutes: 210
+          name: Validate_Arcade_With_Consumer_Repositories
+          displayName: Validate Arcade with Consumer Repositories
+          timeoutInMinutes: 240
           variables:
             - group: Publish-Build-Assets
             - group: DotNetBot-GitHub
             - group: DotNet-Blob-Feed
             - group: DotNet-Symbol-Server-Pats
             - group: DotNet-VSTS-Infra-Access
+          strategy:
+            matrix:
+              ValidateWithRuntime:
+                _azdoOrg: "dnceng"
+                _azdoProject: "internal"
+                _buildDefinitionId: 679
+                _githubRepoName: "runtime"
+                _azdoToken: $(dn-bot-dnceng-build-rw-code-rw)
+                _optionalParameters: "-azdoRepoName 'dotnet-runtime'"
+              ValidateWithRoslyn: 
+                _azdoOrg: "devdiv"
+                _azdoProject: "devdiv"
+                _buildDefinitionId: 8972
+                _githubRepoName: "roslyn"
+                _azdoToken: $(dn-bot-devdiv-build-rw-code-rw)
+                _optionalParameters: "-buildParameters '{\"SkipApplyOptimizationData\":\"true\"}' -pushBranchToGithub"
           steps:
             - checkout: self
               clean: true
             - powershell: eng\validation\build-arcadewithrepo.ps1
-                -azdoOrg "devdiv" 
-                -azdoProject "devdiv"
-                -buildDefinitionId 8972
-                -azdoToken $(dn-bot-devdiv-build-rw-code-rw)
+                -azdoOrg $(_azdoOrg)
+                -azdoProject $(_azdoProject)
+                -buildDefinitionId $(_buildDefinitionId)
+                -azdoToken $(_azdoToken)
                 -githubUser "dotnet-bot"
                 -githubPAT $(BotAccount-dotnet-bot-repo-PAT)
                 -githubOrg "dotnet"
-                -targetRepoName "roslyn"
+                -githubRepoName $(_githubRepoName)
                 -barToken $(MaestroAccessToken)
-                -buildParameters '{"SkipApplyOptimizationData":"true"}' 
+                $(_optionalParameters)
--- a/eng/validation/build-arcadewithrepo.ps1
+++ b/eng/validation/build-arcadewithrepo.ps1
@@ -20,16 +20,29 @@ $ErrorActionPreference = 'Stop'
 . $PSScriptRoot\..\common\tools.ps1
 . $PSScriptRoot\..\common\darc-init.ps1
 
-$arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
-$arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$arcadeSdkPackageName
+$global:arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
+$global:arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$global:arcadeSdkPackageName
+$global:azdoOrg = $azdoOrg
+$global:azdoProject = $azdoProject
+$global:buildDefinitionId = $buildDefinitionId
+$global:azdoToken = $azdoToken
+$global:githubUser = $githubUser
+$global:githubPAT = $githubPAT
+$global:githubOrg = $githubOrg
+$global:githubRepoName = $githubRepoName
+$global:barToken = $barToken
+$global:buildParameters = if (-not $buildParameters) { "" } else { $buildParameters }
+$global:daysOfOldestBuild = if (-not $daysOfOldestBuild) { 3 } else { $daysOfOldestBuild }
+$global:pushBranchToGithub = $pushBranchToGithub
+$global:azdoRepoName = if (-not $azdoRepoName) { "" } else { $azdoRepoName }
 
 # Get a temporary directory for a test root. Use the agent work folder if running under azdo, use the temp path if not.
 $testRootBase = if ($env:AGENT_WORKFOLDER) { $env:AGENT_WORKFOLDER } else { $([System.IO.Path]::GetTempPath()) }
 $testRoot = Join-Path -Path $testRootBase -ChildPath $([System.IO.Path]::GetRandomFileName())
 New-Item -Path $testRoot -ItemType Directory | Out-Null
 
-$minTime = (Get-Date).AddDays(-$daysOfOldestBuild)
-$buildReasonsList = @("batchedCI", "individualCI")
+$global:minTime = (Get-Date).AddDays(-$global:daysOfOldestBuild)
+$global:buildReasonsList = @("batchedCI", "individualCI")
 
 function Get-LastKnownGoodBuildSha()
 {
@@ -37,9 +50,9 @@ function Get-LastKnownGoodBuildSha()
     $headers = Get-AzDOHeaders
     $count = 0
 
-    foreach($reason in $buildReasonsList)
+    foreach($reason in $global:buildReasonsList)
     {
-        $uri = Get-AzDOBuildUri -queryStringParameters "&statusFilter=completed&definitions=${buildDefinitionId}&reasonFilter=${reason}&minTime=${minTime}&`$top=1"
+        $uri = Get-AzDOBuildUri -queryStringParameters "&statusFilter=completed&definitions=${global:buildDefinitionId}&reasonFilter=${reason}&minTime=${global:minTime}&`$top=1"
         $reponse = (Invoke-WebRequest -Uri $uri -Headers $headers -Method Get) | ConvertFrom-Json
         $count += $reponse.count
     }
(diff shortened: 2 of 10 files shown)


eng/validation/build-arcadewithrepo.ps1
```diff
@@ -167,10 +178,14 @@ function Get-Github-RepoAuthUri($repoName)
     "https://${githubUser}:${githubPAT}@github.com/${githubOrg}/${repoName}"
 }
 
+function Get-AzDO-RepoAuthUri($repoName)
```
JohnTortugo: Is this called from many places? If not I'd suggest to inline it.

missymessa:
@JohnTortugo 

> LGTM. Only two questions: 1)can you point to a successful ran with these changes? 2) why are you copying the files in `eng/common` manually ?

1) This was my last test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=575348&view=results

I know it shows that the steps failed, but they were successful in the fact that they kicked off the builds in the other repos after applying the latest Arcade. 

2) I didn't mean to do that. I was testing things locally which meant updating dependencies on my branch, and it updated all of those files :( Is there an easy way to revert that so it's not a part of my check-in?

riarenas:
> I didn't mean to do that. I was testing things locally which meant updating dependencies on my branch, and it updated all of those files :( Is there an easy way to revert that so it's not a part of my check-in?

You're actually downgrading the version and hence you got some merge conflicts. You can run another update-dependencies from the validation channel and your branch should be good.

azure-pipelines.yml
```diff
@@ -209,26 +209,42 @@ stages:
...
             - group: DotNet-Blob-Feed
             - group: DotNet-Symbol-Server-Pats
             - group: DotNet-VSTS-Infra-Access
+          strategy:
+            matrix:
+              ValidateWithRuntime:
+                _azdoOrg: "dnceng"
+                _azdoProject: "internal"
+                _buildDefinitionId: 679
+                _githubRepoName: "runtime"
```
riarenas: Since roslyn is the exception in that it build from GitHub, would it make sense to make the github repository the optional parameter? I don't think we care about the github repo at all when it comes to runtime, or any of the other repos that we're going to validate against.

riarenas: Ah, I see that you're cloning the github repo always. I kind of expected this to clone the azure devops mirror instead since that is the actual repo where the official build runs from. That way you don't have to add a remote, since you're cloning the real repo. This approach seems fine to me as well, if a little roundabout.

There's a very very very small concern where the azdo branch might have different commits than github, and the mirror would be broken in that case, so we'd end up building against the wrong branch, but the chances of that are so small, and all sorts of things would be broken anyways that I think this is fine as well.

azure-pipelines.yml
```diff
@@ -209,26 +209,42 @@ stages:
...
+                _optionalParameters: "-buildParameters '{\"SkipApplyOptimizationData\":\"true\"}' -pushBranchToGithub"
           steps:
             - checkout: self
               clean: true
             - powershell: eng\validation\build-arcadewithrepo.ps1
-                -azdoOrg "devdiv" 
-                -azdoProject "devdiv"
-                -buildDefinitionId 8972
-                -azdoToken $(dn-bot-devdiv-build-rw-code-rw)
+                -azdoOrg $(_azdoOrg)
```
riarenas: Not sure if it makes life simpler for you, but you could just send the github and azdo full repo urls, those include the org and project and you can extract them from there, so that there don't need to be as many params to this.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -195,52 +210,105 @@ function Git-Command($repoName) {
...
 }
 
-## If able to retrieve a build, get the SHA that it was built from
-$sha = Get-LastKnownGoodBuildSha -buildDefinitionId $buildDefinitionId
-
-## Clone the repo from git
-GitHub-Clone $targetRepoName
-
-## Create a branch from the repo with the given SHA. 
+## Global Variables
```
riarenas: These should be marked as $global;<var-name> to make it really clear they are global. it tripped me up in some functions as I didn't know where some of the used variables were coming from.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -195,52 +210,105 @@ function Git-Command($repoName) {
...
+else
+{
+    ## Push branch to AzDO org/project with the official pipeline to build the repo
+    ## make remote, it might already exist if we had to delete it earlier, so wrapping it in a try/catch
+    try
+    {
+        Git-Command $githubRepoName remote add $remoteName $azdoUri
+    }
+    catch
+    {
```
riarenas: won't this swallow legitimate errors? such as auth failures or azdo being down? Does it matter?

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -195,52 +210,105 @@ function Git-Command($repoName) {
...
-## Don't need to do this for Roslyn, but we'll need to do it for Runtime. 
+if($true -eq $pushBranchToGithub)
+{
+    ## Push branch to github
+    Git-Command $githubRepoName push origin HEAD
 
-## Add default channel from that github repo and branch to "General Testing"
-& darc add-default-channel --channel "General Testing" --branch $darcBranchName --repo $darcRepoName --github-pat $githubPAT --password $barToken
+    ## Add default channel from that github repo and branch to "General Testing"
+    & darc add-default-channel --channel "General Testing" --branch $darcBranchName --repo $darcGitHubRepoName --github-pat $githubPAT --password $barToken
```
riarenas: [minor nit] You can take this and LN 299 out of the if so that there's only one darc command. That way if we need to tweak it we only need to do it in one place.

riarenas: Would have the same advantage during deletion, you can skip that whole if.

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -250,27 +318,30 @@ while("completed" -ne (Get-BuildStatus -buildId $buildId))
...
+    else
+    {
+        & darc delete-default-channel --channel "General Testing" --branch $darcBranchName --repo $darcAzDORepoName --azdev-pat $azdoToken --password $barToken
+        Git-Command $githubRepoName push $remoteName --delete $targetBranch
+    }
 }
 catch
 {
-    Write-Warning "Unable to delete default or branch when cleaning up branch"
+    Write-Warning "Unable to delete default channel or branch when cleaning up branch"
```
riarenas:
```suggestion
    Write-Warning "Unable to delete default channel or branch when cleaning up"
```

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -195,52 +210,105 @@ function Git-Command($repoName) {
...
+$githubUri = Get-Github-RepoAuthUri $githubRepoName
+$azdoUri = Get-AzDO-RepoAuthUri $azdoRepoName
+$remoteName = ($azdoOrg + "-" + $azdoRepoName)
 $targetBranch = "dev/" + $githubUser + "/arcade-" + $arcadeSdkVersion
-$githubUri = Get-Github-RepoAuthUri $targetRepoName
-
 $darcBranchName = "refs/heads/" + $targetBranch
-$darcRepoName = "https://github.com/${githubOrg}/${targetRepoName}"
+$darcGitHubRepoName = "https://github.com/${githubOrg}/${githubRepoName}"
+$darcAzDORepoName = "https://${azdoOrg}.visualstudio.com/${azdoProject}/_git/${azdoRepoName}"
```
mmitche: We should use the dev.azure.com name rather than the visualstudio url for this.

azure-pipelines.yml
missymessa: I use `azdoOrg` to construct the URLs for the Build API, too. (And the string I use for the remote name). It would be a which is the lesser evil: parsing a string or more parameters :)

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -195,52 +210,105 @@ function Git-Command($repoName) {
...
-$branchExists = Git-Command $targetRepoName ls-remote --heads $githubUri refs/heads/$targetBranch
+## If able to retrieve a build, get the SHA that it was built from
+$sha = Get-LastKnownGoodBuildSha
+Write-Host "Last Known Good Build SHA: ${sha}"
+
+## Clone the repo from git
+Write-Host "Cloning '${githubRepoName} from GitHub"
+GitHub-Clone $githubRepoName
+ 
+## Check to see if branch exists and clean it up if it does
```
mmitche: Do you need to do cleanup vs. force pushing?

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -195,52 +210,105 @@ function Git-Command($repoName) {
...
-Git-Command $targetRepoName checkout -b $targetBranch $sha
+
+## Create a branch from the repo with the given SHA.
+Git-Command $githubRepoName checkout -b $targetBranch $sha
 
 ## Make the changes to that branch to update Arcade - use darc
-Set-Location $(Get-Repo-Location $targetRepoName)
-& darc update-dependencies --channel ".NET Eng - Validation" --source-repo "arcade" --github-pat $githubPAT --password $barToken
+Set-Location $(Get-Repo-Location $githubRepoName)
+& darc update-dependencies --channel ".NET Eng - Validation" --source-repo "arcade" --github-pat $githubPAT --azdev-pat $azdoToken --password $barToken
```
mmitche: I think this has the possibility of updating to a newer build than is currently running in arcade-validation, if the validation channel gets updated while this build is running. If we have the BAR id of the input build, you could use ```update-dependencies --id```

eng/validation/build-arcadewithrepo.ps1
mmitche: Not for a remote add. There's no contact with external endpoints for that.

eng/validation/build-arcadewithrepo.ps1
riarenas: Oh yeah, you're right. And this is actually a pretty big chance. Would give us results for the wrong arcade build.

mmitche: If you have the arcade version you could look up the BAR build id with get-asset

eng/validation/build-arcadewithrepo.ps1
missymessa: I also use a bunch of the input parameters for the script from AzDO as global variables. Is there a better way of handling those, too? (Do the same thing with the same variable name?)

riarenas: Yeah, anything that is actually global should be marked as such, just makes it easier to know where things are coming from, and avoids errors of overwriting the global when you just meant to overwrite a local var.

eng/validation/build-arcadewithrepo.ps1
missymessa: I have to check to see if the branch exists, and I figured it was easier to blow the whole branch away and start fresh, than trying to repurpose an existing branch. If I don't, I get these error messages: 

```
Running 'git push dnceng-dotnet-runtime dev/dotnet-bot/arcade-5.0.0-beta.20175.3' from D:\a\1ed3w0ky.thy\runtime
To https://dev.azure.com/dnceng/internal/_git/dotnet-runtime
 ! [rejected]                dev/dotnet-bot/arcade-5.0.0-beta.20175.3 -> dev/dotnet-bot/arcade-5.0.0-beta.20175.3 (fetch first)
error: failed to push some refs to 'https://dotnet-bot:***@dev.azure.com/dnceng/internal/_git/dotnet-runtime'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first integrate the remote changes
hint: (e.g., 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
Git exited with exit code: 1
```

eng/validation/build-arcadewithrepo.ps1
missymessa: These commands are taking in different parameters though...

riarenas: Can you populate a $darcRepoName variable with the Github or azdo variable inside the if? Is there another difference?

eng/validation/build-arcadewithrepo.ps1
mmitche: In those cases you could force push. either way works really. Feel free to stick with cleanup.

eng/validation/build-arcadewithrepo.ps1
missymessa: Inlined both of these :)

eng/validation/build-arcadewithrepo.ps1
missymessa: Sorry, no co-author for you :(  I needed to change it in another place, too.

eng/validation/build-arcadewithrepo.ps1
missymessa: Done.

eng/validation/build-arcadewithrepo.ps1
missymessa: > You're actually downgrading the version and hence you got some merge conflicts. You can run another update-dependencies from the validation channel and your branch should be good.

Okay, I think it's where it needs to be. :)

eng/validation/build-arcadewithrepo.ps1
missymessa: I global'd just about everything 😂

eng/validation/build-arcadewithrepo.ps1
missymessa: That works better. I have the conditional assign a $darcRepoName and then use that in the darc command outside the conditional. Also including both GitHub and AzDO tokens in the darc call.

eng/validation/build-arcadewithrepo.ps1
missymessa: Did this. Let me know if the darc call has superfluous parameters or if it's okay the way it is

riarenas: You can drop the repo and channel now that you specify the build

PR #1387, opened 2020-04-01
Adding ASPNETCore repo to Arcade Validation

missymessa:
Other things done in this PR: 
- Removing Roslyn from validation process (they're no longer taking in Arcade via the Latest channel)
- Look up builds on specific branches
- Include manual builds in LKG look-up

Test run of the pipeline with these changes here: https://dnceng.visualstudio.com/internal/_build/results?buildId=583675&view=results

### DIFF
--- a/eng/validation/build-arcadewithrepo.ps1
+++ b/eng/validation/build-arcadewithrepo.ps1
@@ -11,7 +11,8 @@ Param(
   [string] $buildParameters = '',
   [int] $daysOfOldestBuild = 3,
   [switch] $pushBranchToGithub,
-  [string] $azdoRepoName
+  [string] $azdoRepoName,
+  [string] $subscribedBranchName
 )
 
 set-strictmode -version 2.0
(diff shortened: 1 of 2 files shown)


eng/validation/build-arcadewithrepo.ps1
```diff
@@ -210,6 +217,34 @@ function Git-Command($repoName) {
     }
 }
 
+function Get-SubscribedBranch()
```
mmitche, line 220: Is this actually used since the subscribed branch name looks to be passed explicitly?

missymessa: So, it was used. Then, yesterday, I noticed there were two branches subscribed to Latest for both of the repos we're using in this process, so the parameter we pass in was to override the look-up. I'm on the fence about which is better, as consuming repos could subscribe to Latest with a different branch name without us knowing, and then we'd have to do a PR to change the yaml; however, explicitly passing in the branch name allows us to still be able to do the expected look-ups of the LKG and builds as expected, without us having to worry about there being more than one branch subscribed to the Latest channel.

PR #1408, opened 2020-04-07
Made a quick script to promote assets to the Latest channel

missymessa:


missymessa:
> Why is that needed @missymessa ?

Makes it easy for folks who aren't familiar with darc to call this cmdlet with an BAR Build ID. I could just keep the script local on my machine, but I figured it would be more useful to others if I had it in a repo for others to use.

JohnTortugo:
> > Why is that needed @missymessa ?
> 
> Makes it easy for folks who aren't familiar with darc to call this cmdlet with an BAR Build ID. I could just keep the script local on my machine, but I figured it would be more useful to others if I had it in a repo for others to use.

Got it, thanks. The thing is that people usually pushes against having these scripts with just or a few simple commands. I think just calling Darc directly is acceptable in these cases.

/cc @chcosta

chcosta:
I don't really see the benefit of this.  It saves a few key strokes, but only for a very targeted scenario that a very limited number of people are going to exercise and even then, they would have to know this helper script exists..  My preference is that people are forced to learn darc.

chcosta:
I do really like the thought process here though.  You saw a gap and wrote a quick tool, then shared it with the larger group.  I think that thought process is fantastic and should be encouraged.  I just don't see a strong benefit for this particular scenario.

missymessa:
> I don't really see the benefit of this. It saves a few key strokes, but only for a very targeted scenario that a very limited number of people are going to exercise and even then, they would have to know this helper script exists.. My preference is that people are forced to learn darc.

I'll close this and ensure that this gets added to any documentation that I write for this process.

PR #1418, opened 2020-04-16
Output values from build script for telemetry to pick up from AzDO yaml

missymessa:
Part I of #9578. Needed to ensure that we could return this data from the build script so that the (future) API can collect it. 

Verified in https://dnceng.visualstudio.com/internal/_build/results?buildId=604069&view=logs&j=bd570bf9-4a66-5807-2b10-d7a24a0f00a3&t=b801ed86-ba2b-5826-227f-22c29f67055e that the data could be retrieved in subsequent steps (but removed the testing step for the PR).


eng/validation/build-arcadewithrepo.ps1
```diff
@@ -37,6 +37,9 @@ $global:daysOfOldestBuild = if (-not $daysOfOldestBuild) { 3 } else { $daysOfOld
 $global:pushBranchToGithub = $pushBranchToGithub
 $global:azdoRepoName = if (-not $azdoRepoName) { "" } else { $azdoRepoName }
 
+Write-Host "##vso[task.setvariable variable=arcadeVersion;isOutput=true]${global:arcadeSdkVersion}"
+Write-Host "##vso[task.setvariable variable=repo;isOutput=true]${global:githubOrg}/${global:githubRepoName}"
```
JohnTortugo, line 41: NIT: perhaps `qualifiedRepositoryName` would be better.

missymessa: Yep, that's better :)

missymessa:
> LGTM
> 
> I don't remember exactly but I think you only need the isOutput=true if you want the variable to be consumed by another job in YAML.

Yeah, the next part will be writing an API that will take in those variables through another job in the pipeline.

PR #1444, opened 2020-04-27
Add Installer to Arcade Validation pipeline

missymessa:
Arcade Validation pipeline that this ran on: https://dev.azure.com/dnceng/internal/_build/results?buildId=619603&view=results


missymessa:
> LGTM. Not sure about the changes to branchInfo.props as I'm not familiar with how that works.

@mmitche gave me that script, and I verified that it made the change in the branch, so I think it's good :)

PR #1468, opened 2020-05-08
Arcade Validation should only look at product repo latest builds instead of LKGs

missymessa:
We'll use the Latest Build endpoint from AzDO's API to determine if we should validate Arcade against the repo. If the Latest Build result is "partiallySucceeded" or "succeeded", then we'll use the SHA of that build to build Arcade with in that repo. If the result is "cancelled" or "failed", then we won't validate Arcade with that repo.

### DIFF
--- a/eng/validation/build-arcadewithrepo.ps1
+++ b/eng/validation/build-arcadewithrepo.ps1
@@ -171,34 +171,6 @@ function Git-Command($repoName) {
     }
 }
 
-function Get-SubscribedBranch()
-{
-    $targetRepoRegex = "${global:githubRepoName}$"
-    $subscriptions = darc get-subscriptions --target-repo $targetRepoRegex --source-repo "arcade$" --channel ".NET Eng - Latest" --regex --github-pat $global:githubPAT --azdev-pat $global:azdoToken --password $global:bartoken
-
-    if("String" -eq $subscriptions.GetType().Name)
-    {
-        ## If the repo we're testing against doesn't have a subscription to Arcade in .NET Eng - Latest, then we shouldn't test with it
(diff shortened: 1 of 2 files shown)


missymessa:
Validated this code here: https://dnceng.visualstudio.com/internal/_build/results?buildId=635526&view=results

(A couple notes: Removed the DEBUG write host statement for the URI, and changed the "bail out" message to Write-Error instead of Write-Warning so that it will fail the job properly.)

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -45,64 +44,32 @@ $testRootBase = if ($env:AGENT_WORKFOLDER) { $env:AGENT_WORKFOLDER } else { $([S
...
-    ## Have there been any builds in the last $daysOfOldestBuild days? 
+    ## Verified that this API gets completed builds, not in progress builds
     $headers = Get-AzDOHeaders
-    $count = 0
+    $uri = "https://dev.azure.com/${global:azdoOrg}/${global:azdoProject}/_apis/build/latest/${global:buildDefinitionId}?branchName=${global:subscribedBranchName}&api-version=5.1-preview.1"
+    $response = (Invoke-WebRequest -Uri $uri -Headers $headers -Method Get) | ConvertFrom-Json
 
-    foreach($reason in $global:buildReasonsList)
+    ## Is it successful or partially successful? Then use that as the foundation for our branch. 
+    ## How do we want to handle cancelled builds?
```
riarenas, line 55: I think we already have the answer to this question in the code below: we handle them as not successful

missymessa: I'll take that note to self out of there 😂

missymessa:
Looks like Arcade Validation is looking at the latest completed build instead of looking for a LKG in the last three days. Closing.

PR #1577, opened 2020-06-15
Script to clean up old branches on bellwether repos

missymessa:
Every time validation for Arcade fails on a bellwether repo, we retain the branch for investigative purposes. This script will allow us to programmatically clean up old branches (except whatever is current). It uses the same functionality that can be found in the script to create the branches and run the builds used for the validation-for-promotion pipeline. 

After this is merged, I plan to use it in this pipeline: https://dnceng.visualstudio.com/internal/_build?definitionId=864&_a=summary and run it on a weekly schedule.


missymessa:
Validated that this does what it needs to here: https://dnceng.visualstudio.com/internal/_build/results?buildId=688615&view=results

missymessa:
ping @JohnTortugo and @riarenas :)

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,98 @@
...
+$global:githubRepoName = $githubRepoName
+$global:barToken = $barToken
+$global:azdoRepoName = if (-not $azdoRepoName) { "" } else { $azdoRepoName }
+
+# Get a temporary directory for a test root. Use the agent work folder if running under azdo, use the temp path if not.
+$testRootBase = if ($env:AGENT_WORKFOLDER) { $env:AGENT_WORKFOLDER } else { $([System.IO.Path]::GetTempPath()) }
+$testRoot = Join-Path -Path $testRootBase -ChildPath $([System.IO.Path]::GetRandomFileName())
+New-Item -Path $testRoot -ItemType Directory | Out-Null
+
+function GitHub-Clone($repoName)
```
JohnTortugo, line 36: Looks like there are some duplicated functions in this script and in `build-arcadewithrepo.ps1` - e.g., GitHub-Clone, Git-Command. I think it would be best to  extract them and put in a shared/common file.

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,98 @@
...
+        Write-Host "Running 'git $gitParams' from $(Get-Location)"
+        $commandOutput = & git @gitParams; if ($LASTEXITCODE -ne 0) { throw "Git exited with exit code: $LASTEXITCODE" } else { $commandOutput }
+        $commandOutput
+    }
+    finally {
+        Pop-Location
+    }
+}
+
+## Global Variables
```
JohnTortugo, line 63: I'd prefer if all global variables were defined in the same region of the script - at the top.

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,98 @@
...
+        $branchName = ($remoteBranch -split "`t")[1]
+        try
+        {
+            Write-Host "Delete default channel and branch for branch named '${branchName}'"
+            & darc delete-default-channel --channel "General Testing" --branch $branchName --repo $global:darcAzDORepoName --azdev-pat $global:azdoToken --password $global:bartoken
+            Git-Command $global:githubRepoName push $remoteName --delete $branchName
+        }
+        catch
+        {
+            Write-Warning "Unable to delete default channel or branch when cleaning up"
```
JohnTortugo, line 95: Is usually helpful to say what value caused the exception to be triggered. In this case what was the branch name and repo for instance. Also, maybe you can print "$_" - it may provide some useful info during an investigation?!

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,98 @@
...
+Git-Command $global:githubRepoName remote add $remoteName $global:azdoUri
+
+$remoteBranches = @()
+Write-Host "Getting remote branches for '${global:githubRepoName}' on Azure DevOps"
+$remoteBranches = Git-Command $global:githubRepoName ls-remote --heads $global:azdoUri
+
+foreach($remoteBranch in $remoteBranches)
+{
+    ## Delete all of the old branches except for any that match the current version of Arcade being validated
+    if(($remoteBranch -match ".*($global:branchNamePrefix).*") -and -not ($remoteBranch -match ".*($global:lastBranch).*"))
```
JohnTortugo, line 84: @missymessa - I'm quite concerned about what an error calling this script (or in the script.. RegExp!) would do. I don't see anything wrong but the potential effects if something goes wrong is pretty big. Can you make it safer to use this script?

missymessa: I could use `-like` instead of `-match` and get the same results. Do you think that would be better? What other things should I do here to make the script safer?

azure-pipelines-clean-up.yml
```diff
@@ -0,0 +1,44 @@
+stages:
```
riarenas, line 1: I think this can just be named clean-up-branches.yml or something. azure-pipelines.yml is just the default name that azure devops looks for when you're trying to create a new pipeline, so if we're not calling it exactly that, might as well just call it for what it does? current name kind of implies you're cleaning up azure pipelines

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,98 @@
...
+        $commandOutput = & git @gitParams; if ($LASTEXITCODE -ne 0) { throw "Git exited with exit code: $LASTEXITCODE" } else { $commandOutput }
+        $commandOutput
+    }
+    finally {
+        Pop-Location
+    }
+}
+
+## Global Variables
+$global:githubUri = "https://${global:githubUser}:${global:githubPAT}@github.com/${global:githubOrg}/${global:githubRepoName}"
```
riarenas, line 64: [nit] should githubUser just be `user` since it's used for both github and azdo? looks like a bug that you'd use the githubUser for the azdoUri

riarenas:
I think this code could be more readable, require less setup, and be less complex if we used the Azure Devops APIs instead of setting up a local git. The JSON responses from the API (which can be filtered to get exactly what we want anyways, preventing scary situations where we delete are much simpler to parse than the freeform console output from the git executable.

I think this approach works though, so I won't block the PR on that.

eng/validation/remove-oldbranches.ps1
JohnTortugo: I think I was more raising a concern than calling for something specific to be implemented.

missymessa: @riarenas pointed out that AzDO has an API that allows you to filter on branch names, so I'll give that a try to see if it works for this script rather than parsing results from git commands: https://docs.microsoft.com/en-us/rest/api/azure/devops/git/refs/list?view=azure-devops-rest-5.1#refs-heads-that-contain-a-word

azure-pipelines-clean-up.yml
```diff
@@ -0,0 +1,44 @@
...
+  jobs:
+    - template: /eng/common/templates/job/job.yml
+      parameters:
+        name: Clean_Up_Arcade_Branches_On_Consumer_Repositories
+        displayName: Clean Up Arcade Branches On Consumer Repositories
+        timeoutInMinutes: 240
+        pool: 
+          vmImage: vs2017-win2016
+        variables:
+          - group: Publish-Build-Assets
```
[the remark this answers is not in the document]
missymessa: Nope! I cleaned them up here and in the original yaml for validating Arcade for promotion.

missymessa:
Validated the changes on this run: https://dnceng.visualstudio.com/internal/_build/results?buildId=694706&view=results  

Granted, nothing got deleted because I previously deleted them while testing the changes in the script 😂 

But I was able to verify that ASPNETCore still retained the branch with the current version (it had failed originally, so the validation-for-promotion script didn't clean it up).

eng/validation/remove-oldbranches.ps1
missymessa: Changed it to `user`. GitHub references weren't needed for the refactored script anyway :)

eng/validation/remove-oldbranches.ps1
missymessa: Removed this part of the code, but still ensured that any other messages contained more information for debugging purposes.

eng/validation/remove-oldbranches.ps1
missymessa: Removed this code.

eng/validation/remove-oldbranches.ps1
missymessa: Cleaned up the unnecessary variables and moved them up to the top-ish of the script :)

eng/validation/remove-oldbranches.ps1
missymessa: Refactored this script to make use of AzDO's API to look up branches and delete them. Much nicer code than all the git commands.

azure-pipelines-clean-up.yml
missymessa: Renamed the yaml to something that won't get your hopes up on what it does. :)

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,120 @@
...
+  [Parameter(Mandatory=$true)][string] $azdoRepoName
+)
+
+set-strictmode -version 2.0
+$ErrorActionPreference = 'Stop'
+
+. $PSScriptRoot\..\common\tools.ps1
+. $PSScriptRoot\..\common\darc-init.ps1
+
+# Get a temporary directory for a test root. Use the agent work folder if running under azdo, use the temp path if not.
```
riarenas, line 16: I think these aren't being used anymore?

eng/validation/remove-oldbranches.ps1
```diff
@@ -0,0 +1,120 @@
...
+            $oldObjectId = $remoteBranch.objectId
+            $json = @{
+                "name"=$branchName;
+                "oldObjectId"=$oldObjectId;
+                "newObjectId"="0000000000000000000000000000000000000000"
+            }
+            
+            $jsonBodyArray += $json
+
+            Write-Host "Sending '${branchName}' to AzDO API to be deleted"
```
riarenas, line 92: It looks like this log statement should be somewhere around LN 97

missymessa: I was trying to figuring out wording here to say "Identified this branch as something that needs to be deleted, so I'm configuring the body of the API request to take in this branch." :)

riarenas: ahh I see what you're doing here. Still a little confusing as we're not deleting them yet, and might cause some confusion when examining the logs if something seems off.

missymessa: I'll try to find better wording for this log.

riarenas: How about something like:

```
Gathering branches to be deleted:
Found branch <blah>. Deleting default channel
```
loop finishes
```
Deleting branches...

PR #1629, opened 2020-07-08
Stop producing package for Validation.Tests

riarenas:
This is preparation for when we update the version of the Microsoft.NET.Test.Sdk in Arcade. Part of https://github.com/dotnet/arcade/issues/5291

### DIFF
--- a/src/Validation/tests/Validation.Tests.csproj
+++ b/src/Validation/tests/Validation.Tests.csproj
@@ -2,6 +2,5 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
-    <IsPackable>true</IsPackable>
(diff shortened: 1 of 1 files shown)


garath:
How would you feel about adding some text in the Arcade documentation? I think the [FAQ](https://github.com/dotnet/arcade/blob/master/Documentation/FAQ.md) is a good candidate to describe all the details about this and also identify the few packages that aren't managed by subscription.

riarenas:
I might be missing the value or what you mean would be a good candidate for documentation?

I don't think it makes a lot of sense to make a doc with a distinction about packages that aren't managed by dependency flow and those that aren't, especially since that information is already available in the version.details.xml file.

I also don't think the FAQ is a good place for transient reactions to version updates.

PR #1684, opened 2020-08-05
Add Setup private feeds credentials during official build stage

riarenas:
https://github.com/dotnet/arcade/issues/5788 and some validation for https://github.com/dotnet/arcade/issues/5890

* Add build steps to run the SetupNugetSources scripts in linux and windows builds.
* Add a fake source that the scripts should not re-enable as it will cause nuget to die
* Add a real source that the logs should show is re-enabled.

Some additional validation to make sure the nuget.config file looks as expected should be added to complete the validation.

Test build: https://dnceng.visualstudio.com/internal/_build/results?buildId=758611&view=results


riarenas:
We need a valid PAT, or nuget will barf when it's unable to load the feed.

riarenas:
We don't need a valid PAT right now since the feed is going to stay disabled, but after the incoming change, it's required, as it will be enabled, and nuget will check the feed for packages (it won't find any of the ones we need, but it will still try to access the feed)

PR #1761, opened 2020-08-31
Add all files in working tree before commiting

riarenas:
fixes https://github.com/dotnet/arcade/issues/6075

`git commit -am` does not include new files. so run a git add beforehand.

Test build (that will also serve as a promotion validation for today): https://dnceng.visualstudio.com/internal/_build/results?buildId=794933&view=results


riarenas:
All bellwether repos are now building correctly. Any errors in the validation would not be related to this, so merging.

PR #1873, opened 2020-10-12
Promote build automatically to .net 5 eng

riarenas:
https://github.com/dotnet/arcade/issues/6390

This change makes it so the official build will promote the arcade build to the .net 5 eng channel automatically.

It takes advantage of the --skip-assets-publishing flag because we don't need to republish any assets.

I also ported the change from https://github.com/dotnet/arcade-validation/pull/1869 so that private builds of the release/5.0 branch don't do the "validate publishing" stage.

Test build (this added the missing build to the channel even) https://dnceng.visualstudio.com/internal/_build/results?buildId=849418&view=results

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -224,85 +224,47 @@ stages:
         -TsaRepositoryName "Arcade-Validation"
         -TsaCodebaseName "Arcade-Validation"
         -TsaPublish $True'
-  - stage: Validate_Publishing
-    displayName: Validate Publishing
-    jobs: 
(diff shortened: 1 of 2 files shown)


mmitche (approved):

missymessa:
Do we want to make the same changes for update-channel.ps1 in master as we're doing for this branch?

riarenas:
>Do we want to make the same changes for update-channel.ps1 in master as we're doing for this branch?

Maybe if we're ever going to turn on automatic promotion in master again? I don't think I should be doing it as part of this issue though.

missymessa (approved):
LGTM shipit

PR #1988, opened 2020-12-07
Add a small infrastructure for running E2E tests in arcade-validation

mmitche:
Arcade-validation today is sorta a single test case. Build repo and ensure
it passes. This change adds a small test framework that allows for on-the-fly
creation of small new arcade repo layouts, building them, and checking results.

It uses the following:
- A optional common repo layout that avoids downloading the same packages and zips over and over
  over again. Instead, a small repo is created, restored, and then that repo .dotnet/.packages dirs
  are used in other tests.
- A set of standard set-up methods in a builder to create typical basic repo layouts.

The main goal of this is to avoid checking in as many files as possible. It only
copies the eng/common dir from the parent arcade-validation repo.

I added 5 test cases. 1 basic one and 4 focusing on error behavior in
signing. https://github.com/dotnet/arcade/issues/6367. One of these actually
fails (the post build sign one) today in these tests because they can't actually publish to AzDO (permissions of course)
and the error condition for an empty post build sign list lives in the wrong place today.

https://github.com/dotnet/arcade/pull/6656


src/Validation/tests/RepoTests.cs
```diff
@@ -0,0 +1,142 @@
...
+                        .SdkCsproj(
+                            targetFramework: "net5.0",
+                            outputType: "Exe")
+                        .PropertyGroup()
+                        .Property("IsPackable", "true"),
+                    "src/FooPackage/FooPackage.csproj");
+                await builder.AddSimpleCSFile("src/FooPackage/Program.cs");
+
+                builder.Build("-configuration", "Release", "-restore", "-sign", "-ci", "-projects", "src/FooPackage/FooPackage.csproj")
+                    .Should().Throw<Exception>().WithMessage("*error : List of files to sign is empty. Make sure that ItemsToSign is configured correctly*");
```
missymessa, line 88: Is there another way we can identify this error instead of string compare just in case the string changes in the future? (same for the other places we do this check)

missymessa (approved):
I love this. Very exciting!

LGTM shipit

src/Validation/tests/RepoTests.cs
mmitche: Not that I know of, but i'm open to suggestions. This is really checking for the msbuild error message. So we're checking against what the user should see.

missymessa: I suppose if this error message changes, we'll get an error on this test and can address it then. Was hoping that maybe we could look for a specific exception type, but not sure if that's something we can do with MSBuild stuff. No big deal either way.

PR #2077, opened 2021-01-28
Switch to using a PAT that works for both dnceng and devdiv

MattGal:
See https://github.com/dotnet/core-eng/issues/11993
It woudl appear the Installer repo ends up indirectly cloning repos from devdiv.vs now.  

Since we have a PAT that works for both projects .NET Core development runs in, may as well use that for all as any repo could do this in the future.

### DIFF
--- a/azure-pipelines-bellwether-repos.yml
+++ b/azure-pipelines-bellwether-repos.yml
@@ -20,21 +20,21 @@ stages:
               _azdoProject: "internal"
               _buildDefinitionId: 679
               _githubRepoName: "runtime"
-              _azdoToken: $(dn-bot-dnceng-build-rw-code-rw)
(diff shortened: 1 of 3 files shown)


missymessa:
You also need to make changes to the https://github.com/dotnet/arcade-validation/blob/master/azure-pipelines-bellwether-repos.yml too

missymessa (approved):
LGTM shipit

riarenas:
>It would appear the Installer repo ends up indirectly cloning repos from devdiv.vs now.

Yep, this is because Nuget recently switched to producing their builds from an internal-only repository and we haven't fixed up maestro to understand the mapping between that repo and github nuget. So the coherency algorithm tries to access the devidv repo. See https://github.com/dotnet/arcade/issues/6813 if more context is needed

PR #2082, opened 2021-01-29
Handle scenario of no darc updates for a given repo

MattGal:
Since this isn't necessarily indicative of anything wrong, just print warning and exit.  Also print output when we have git errors.

https://github.com/dotnet/core-eng/issues/11993

### DIFF
--- a/eng/validation/build-arcadewithrepo.ps1
+++ b/eng/validation/build-arcadewithrepo.ps1
@@ -45,6 +45,8 @@ $testRootBase = if ($env:AGENT_WORKFOLDER) { $env:AGENT_WORKFOLDER } else { $([S
 $testRoot = Join-Path -Path $testRootBase -ChildPath $([System.IO.Path]::GetRandomFileName())
 New-Item -Path $testRoot -ItemType Directory | Out-Null
 
+Write-Output "Working inside $testRoot"
+
 function Get-LatestBuildSha()
 {
     ## Verified that this API gets completed builds, not in progress builds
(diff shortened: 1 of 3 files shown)


MattGal:
https://dnceng.visualstudio.com/internal/_build/results?buildId=971547&view=results

eng/validation/build-arcadewithrepo.ps1
```diff
@@ -202,7 +204,15 @@ $barBuildId = ([regex]"\d+").Match($barBuildIdString).Value
...
 Set-Location $(Get-Repo-Location $global:githubRepoName)
-& $darc update-dependencies --id $barBuildId --github-pat $global:githubPAT --azdev-pat $global:azdoToken --password $global:bartoken
+
+$darcOutput = & $darc update-dependencies --id $barBuildId --github-pat $global:githubPAT --azdev-pat $global:azdoToken --password $global:bartoken
+Write-Output $darcOutput
+if ($darcOutput -eq "Found no dependencies to update.")
+{
+   ## Log a warning so this isn't hidden, but it's not necessarily a problem either, just exit.
+   Write-Host "##vso[task.logissue type=warning]Darc returned no dependencies to update for this repo"
+   exit 0
```
missymessa, line 214: Let's have this return a failure. If the dependencies fail to update in this scenario, then we have an issue somewhere that needs to be fixed, and this particular pipeline will end up returning a false positive on a passing build with the latest Arcade version.

MattGal: OK cool.  I'll promote it to error too.

MattGal: Sure, also kicked off this version here; https://dnceng.visualstudio.com/internal/_build/results?buildId=971553&view=results

missymessa (approved):
LGTM shipit

PR #2155, opened 2021-03-02
SDK updated to 6.0, so we have to specify to use 5.0 runtime

missymessa:


riarenas (approved):

jonfortescue (approved):
A+

missymessa:
Arg, I did it again. Thanks @riarenas for fixing my assignee/reviewer assignments.

MattGal (approved):

ViktorHofer (commented):
So I presume you don't want to update all your test applications to target net6.0?

missymessa:
> So I presume you don't want to update all your test applications to target net6.0?

That would've worked, but we wanted to make sure that we had a solution in this repo that wouldn't have required it (which means, it would've required other repos to make the same change).

ViktorHofer:
When upgrading the SDK to a different major versions, repos are supposed to be required to updated their tfms so that they actually target and run on the latest and greatest of the stack. Adding a 5.0 runtime to be present is well is only a temporary mitigation.

riarenas:
I don't recall seeing such an expectation on our policy docs. I don't think it's entirely reasonable to expect consumer repos to all in update their tfms for preview 1.

ViktorHofer:
Using a 6.0 SDK doesn't require tfms to be updated to net6.0. Changes are only necessary in places where the app is invoked without a compliant roll forward policy. In this case, a test's runtimeconfig.json was used to invoke xunit.console and as the 5.0.0 runtime wasn't available, xunit.console couldn't be invoked via dotnet.

If the command that invoked xunit.console would have had a rollforward setting (which is possible via dotnet exec), this issue wouldn't have happened.

> I don't recall seeing such an expectation on our policy docs. I don't think it's entirely reasonable to expect consumer repos to all in update their tfms for preview 1.

On the contrary, I wouldn't expect an SDK upgrade forcing consumers to include a previous shipped runtime for their test execution.

riarenas:
Repos are not forced to add a runtime as a way to unblock, but I think it's important that repositories are given the option while we're flowing preview builds.

mmitche:
@ViktorHofer I don't think we'll see this forcing downstream repos to alter their global.jsons. The issue is specific to this repo. We could upgrade this repo's tfms, though I don't see a ton of value in it. All the downstream repos have already upgraded their SDKs to 6.0 to target the 6.0 tfm, including their test execution. E.g.:

https://github.com/dotnet/aspnetcore/blob/main/global.json

ViktorHofer:
> We could upgrade this repo's tfms, though I don't see a ton of value in it.

Which probably leads to the question, what is this repo supposed to test and on which runtime?

> All the downstream repos have already upgraded their SDKs to 6.0 to target the 6.0 tfm, including their test execution. E.g.:

Which is a good thing, for both dogfooding and sourcebuild.

missymessa:
Since it sounds like the other Arcade-consuming repos have updated their TFMs, I'll make this change in the Arcade Validation repo when I'm in there for another issue (in the near future) :)

PR #2167, opened 2021-03-09
Include automated promotion of Arcade after Arcade Validation

missymessa:
**NOTE**: This PR, if/when approved, will be merged after sign-off with management.


mmitche (approved):

MattGal (approved):

epananth (approved):

riarenas (approved):

missymessa:
I forgot to include the check to ensure that the bellwether repos are green! I will update the PR with that functionality.

missymessa:
Hey folks, thanks for your super quick approvals previously, but since I've pushed up some new logic, I'm going to ask for a re-review. Thanks :)

eng/validation/update-channel.ps1
```diff
@@ -24,65 +26,118 @@ function Get-AzDO-Build([string]$token, [int]$azdoBuildId) {
...
+    ## Verified that this API gets completed builds, not in progress builds
+    $headers = Get-AzDOHeaders
+    $uri = "https://dev.azure.com/$($repoData.azdoOrg)/$($repoData.azdoProject)/_apis/build/latest/$($repoData.buildDefinitionId)?branchName=$($repoData.subscribedBranchName)&api-version=5.1-preview.1"
+    $response = (Invoke-WebRequest -Uri $uri -Headers $headers -Method Get) | ConvertFrom-Json
+
+    ## Report non-green repos for investigation purposes. 
+    if(($response.result -ne "succeeded") -and ($response.result -ne "partiallySucceeded"))
+    {
+        Write-PipelineTaskError -message "The latest build on '$($repoData.subscribedBranchName)' branch for the '$($repoData.githubRepoName)' repository was not successful." -type "warning"
+        Write-Host "##vso[task.complete result=SucceededWithIssues;]"
```
riarenas, line 47: [nit] 

This will only matter if we try to move this function to a common place where other scripts would use it, but  I wonder if it makes sense to move this logging command to the caller instead of this function after it sees that the return value was a false. It seems like an unexpected side effect that calling a Get function would mark my build as succeeding with issues.

If we move it, it would also mean we only send the logging command once instead of N times for each repo that isn't gree

riarenas (approved):
LGTM!

eng/validation/update-channel.ps1
missymessa: Good point. I moved it out to the else case in the conditional that checks the results of the bellwethers.

riarenas (approved):

epananth (approved):

PR #2219, opened 2021-03-31
Fix logging for validation

adiaaida:
The logging functions require $ci to be $true. Setting this variable causes the lines to output with the necessary vso headers

### DIFF
--- a/eng/validation/update-channel.ps1
+++ b/eng/validation/update-channel.ps1
@@ -7,6 +7,7 @@ Param(
   [string] $githubToken
 )
 
+$ci = $true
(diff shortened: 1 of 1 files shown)


missymessa:
Yep!  I just want to make sure that the warning message shows up in the list of warnings/errors on AzDO, so that it'll be picked up by Build Monitor :)

adiaaida:
Run will be here: https://dev.azure.com/dnceng/internal/_build/results?buildId=1066111&view=results

missymessa:
It did the thing!

missymessa (approved):
LGTM shipit

PR #2249, opened 2021-04-14
Skip component governance detection where not needed

premun:
Related to https://github.com/dotnet/core-eng/issues/12816


missymessa (approved):
I assume this is fine, but I'm not super familiar with this task.

PR #2301, opened 2021-05-10
Add tests exercising UseDotNetCertificate

mmitche:
Add a few tests for this feature, which replaces Microsoft400 with MicrosoftDotNet500 when the flag is set. Defaults to false.
Ensure that this works on two paths:
- Post build signing
- Regular signing

Necessary to have tests for both these cases because the places where the certificate replacement is different for PBS and in-build signing


missymessa (approved):
LGTM shipit

PR #2365, opened 2021-06-11
Updated xcopy-msbuild to 16.10.0-preview2 to fix signing issue

jakubstilec:
https://github.com/dotnet/arcade/issues/7494


lpatalas (approved):
I'm not an expert on this but apparently it fixes the issue :)

riarenas:
@jakubstilec since these changes were made to the eng/common folder, they need to be made in Arcade. If we leave things like this we will break every repo because when arcade changes flow to consumer repos they won't include these script changes. In fact, the very next merge to arcade will overwrite them here.

PR #2509, opened 2021-08-31
Always use 1ES Pools

ulisesh:
By default we were using NetCorePublic-Pool and NetCoreInternal-Pool for internal builds, changing them to be 1ES pools


MattGal (approved):
Approving assuming you've successfully run this build on those agents.

PR #2534, opened 2021-09-13
Use release/6.0 branch to promote to .Net 6 Eng Channel

epananth:
Release/6.0 branch should be used instead of main to promote to .Net 6 Eng Channel

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -232,7 +232,7 @@ stages:
         -TsaRepositoryName "Arcade-Validation"
         -TsaCodebaseName "Arcade-Validation"
         -TsaPublish $True'
-  - ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
+  - ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/release/6.0') }}:
     - stage: Validate_Publishing
       displayName: Validate Publishing
       jobs:
@@ -263,15 +263,15 @@ stages:
                 -azdoProject "internal"
                 -barToken $(MaestroAccessToken)
                 -githubPAT $(BotAccount-dotnet-bot-repo-PAT)
-    - stage: Promote_Arcade_To_Latest
-      displayName: Promote Arcade to '.NET Eng - Latest' channel
+    - stage: Push_to_net6_eng_channel
+      displayName: Promote Arcade to '.NET 6 Eng' channel
       dependsOn:
         - Validate_Publishing
       jobs:
       - template: /eng/common/templates/job/job.yml
         parameters:
-          name: Promote_Arcade_To_Latest
-          displayName: Promote Arcade to '.NET Eng - Latest' channel
+          name: Push_to_net6_eng_channel
+          displayName: Promote Arcade to '.NET 6 Eng' channel
           timeoutInMinutes: 180
           variables:
             - group: Publish-Build-Assets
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -263,15 +263,15 @@ stages:
                 -azdoProject "internal"
                 -barToken $(MaestroAccessToken)
                 -githubPAT $(BotAccount-dotnet-bot-repo-PAT)
-    - stage: Promote_Arcade_To_Latest
-      displayName: Promote Arcade to '.NET Eng - Latest' channel
+    - stage: Push_to_net6_eng_channel
+      displayName: Promote Arcade to '.NET 6 Eng' channel
```
MattGal, line 267: This looks right, but there seems to be a mismatch between "promote" and "push"?

MattGal (approved):

riarenas:
This is needed, but I think it's not enough. I think we need to make changes to https://github.com/dotnet/arcade-validation/blob/release/6.0/eng/validation/update-channel.ps1 so that it looks more like the release/5.0 version instead of the main version of the file. 

For release branches we don't check on the bellwether repo state, so I think the script shouldn't check the latest main build for each pipeline, and instead promote automatically.

azure-pipelines.yml
epananth: I followed what we have for release/5.0 branch to keep that consistent..  Changed it back to promote

epananth:
Sure will update that file.. Thank you !

eng/validation/update-channel.ps1
```diff
@@ -51,65 +51,35 @@ function Get-LatestBuildResult([PSObject]$repoData)
...
-$bellwetherRepos = @($runtimeRepo, $aspnetcoreRepo, $installerRepo)
-
 $arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
 $arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$arcadeSdkPackageName
 $getAssetsApiEndpoint = "$maestroEndpoint/api/assets?name=$arcadeSdkPackageName&version=$arcadeSdkVersion&api-version=$apiVersion"
 $headers = Get-Headers 'text/plain' $barToken
 
 try {
-    # Validate that the "bellwether" repos (runtime, installer, aspnetcore) are green on their main branches
-    $results = ($bellwetherRepos | ForEach-Object { Get-LatestBuildResult -repoData $_ })
```
riarenas, line 85: You are removing the usage for Get-LatestBuildResult, we should delete the function as well.

epananth: lol, you beat me to it.. I just did :)

eng/validation/update-channel.ps1
```diff
@@ -34,82 +34,35 @@ function Get-AzDOHeaders()
     return $headers
 }
 
-function Get-LatestBuildResult([PSObject]$repoData)
-{
-    ## Verified that this API gets completed builds, not in progress builds
-    $headers = Get-AzDOHeaders
```
riarenas, line 40: I think same thing with Get-AzDOHeaders and Get-AzDO-Build. Looks like this was their only usage? 
potentially other functions in this file are in the same boat. . Let's only keep code that is actually being used.

epananth: I was following release/5.0 and just realized that has zero references. I will remove that from both release/6.0 and release/5.0

riarenas (approved):

PR #2553, opened 2021-09-21
[release/6.0] Migrate to 1ES hosted pools

lpatalas:
Tracking issue: https://github.com/dotnet/core-eng/issues/14276

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,16 +8,16 @@ variables:
     - group: SDL_Settings
   - ${{ if and(eq(variables.PoolProvider, ''), eq(variables['System.TeamProject'], 'public')) }}:
     - name: PoolProvider
-      value: NetCorePublic-Pool
+      value: NetCore1ESPool-Svc-Public
   - ${{ if and(ne(variables.PoolProvider, ''), ne(variables['System.TeamProject'], 'public')) }}:
     - name: PoolProvider
-      value: NetCoreInternal-Pool
+      value: NetCore1ESPool-Svc-Internal
   - ${{ if and(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')) }}:
     - name: PoolProvider
-      value: NetCorePublic-Int-Pool
+      value: NetCore1ESPool-Svc-Public-Int
   - ${{ if and(ne(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')) }}:
     - name: PoolProvider
-      value: NetCoreInternal-Int-Pool
+      value: NetCore1ESPool-Svc-Internal-Int
 
 trigger:
   batch: true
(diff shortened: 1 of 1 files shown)


jonfortescue (approved):

MattGal (approved):

riarenas:
@lpatalas 
```
##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider BadRequest
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider BadRequest
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider BadRequest
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider BadRequest
,##[warning]There was a failure in sending the provision message: Unexpected response code from remote provider BadRequest
,##[error]Provisioning request delayed or failed to send 5 time(s). This is over the limit of 3 time(s).
Pool: NetCoreInternal-Pool
```

Does this mean it didn't really try to use the 1es pools, and ended up with a mix of things that just don't work?

lpatalas:
@riarenas From what I see the pool is set to `NetCoreInternal-Pool` in pipeline variables in AzDO. I will change it to the new one. What's the reason behind having the pool overridden in the pipeline?

riarenas:
Ooh, good find. I have no explanation for that and I don't think it's correct that we're overriding it.

riarenas:
In fact, changing the variable for the pipeline would break main wouldn't it? I think we should just unlink that variable from the pipeline and just let the YAML do things. If somehow the YAML is not passing the variable correctly, fix that across branches.

lpatalas:
Ok, I will prepare another PR that will ignore that global variable in the yaml.

riarenas:
I don't think we need a PR, I mean removing the variable from the pipeline variables, so that it can take the value set by the YAML. Let me give that a quick try and get back to you whether we need to make any yaml changes

riarenas:
Sadly it's what I feared. After removing the pipeline variable from the UI, the errors I get are:

##[error]The pipeline is not valid. Could not find a pool with name $(PoolProvider). The pool does not exist or has not been authorized for use. For authorization details, refer to https://aka.ms/yamlauthz.

Which means that this block: https://github.com/dotnet/arcade-validation/blob/a405bb7df58b7fbe0ca420c26e79aa979e900974/azure-pipelines.yml#L10-L20

Is not doing what we expected. It also means that without the variable, things are pretty broken without it. Basically we let a hack of adding the pool to the pipeline variables go on for too long. I re-added the variable, and I will file an issue to fix the YAML.

riarenas:
The main problem now is that pipeline variables are not set per branch, but per pipeline, so we have no way to tell main to go to the R&D provider but have release go to the servicing provider. 

This is not good.

lpatalas:
From what I see the release/5.0 branch succeeded on the new pools. So I think we have to do the same change that Ulises made here: https://github.com/dotnet/arcade-validation/pull/2509/files.

riarenas:
Seems reasonable

lpatalas:
I've cherry-picked Ulises change here: https://github.com/dotnet/arcade-validation/pull/2570

PR #2570, opened 2021-09-24
Always use 1ES Pools

lpatalas:
We have global `PoolProvider` variable defined on the pipeline. This causes issues when pools are changed to 1ES hosted ones. See the other PR for discussion: https://github.com/dotnet/arcade-validation/pull/2553.


riarenas:
LGTM, I also queued https://dev.azure.com/dnceng/internal/_build/results?buildId=1384096&view=results with these same changes to see whether this chooses the right pools, and it seems like it did

riarenas:
I will merge this once I'm back from my appointment. Thanks for figuring this out @lpatalas

PR #2525, opened 2021-09-08
Use 1ES R&D pools

ulisesh:

PR #2590, opened 2021-10-01
Revert "Use R&D pools (#2525)"

lpatalas:
Staging CloudTest images have `-Int` suffix but we forgot to take it into account when migrating the pipeline. Also we need to increase max scale on staging pools. This PR temporarily reverts to the old pools to unblock the promotion.

Issue: https://github.com/dotnet/arcade/issues/7987

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,16 +8,16 @@ variables:
     - group: SDL_Settings
   - ${{ if eq(variables['System.TeamProject'], 'public') }}:
     - name: PoolProvider
-      value: NetCore1ESPool-Public
+      value: NetCorePublic-Pool
   - ${{ if ne(variables['System.TeamProject'], 'public') }}:
     - name: PoolProvider
-      value: NetCore1ESPool-Internal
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -8,16 +8,16 @@ variables:
     - group: SDL_Settings
   - ${{ if eq(variables['System.TeamProject'], 'public') }}:
     - name: PoolProvider
-      value: NetCore1ESPool-Public
+      value: NetCorePublic-Pool
```
riarenas, line 11: I don't think we need to revert the non-scheduled runs, those are the only ones I've seen failing, because those are the only ones that run in int. I'm fine with the full revert if that's easier for you though.

riarenas (approved):

azure-pipelines.yml
lpatalas: It would be more tricky to revert it partially because the image names differ between Helix and 1ES pools.

ulisesh (approved):
LGTM

PR #2619, opened 2021-10-11
Migrate to 1ES pools

lpatalas:
We have renamed staging 1ES images to match production names. Now we can once again apply the migration to 1ES  hosted pools.

Tracking issue: https://github.com/dotnet/arcade/issues/7987


ulisesh (approved):
LGTM

PR #2651, opened 2021-10-19
[main] Use ubuntu-latest vm image during official builds

ghost:
Hosted ubuntu 1604 stopped being valid. Moving us to ubuntu-latest. If ubuntu-latest stops meeting our needs for the build, it will fail instead of silently showing up as cancelled. 

https://github.com/dotnet/arcade/issues/8066

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -112,7 +112,7 @@ stages:
         pool:
           ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}:
             name: $(PoolProvider) # This is a queue-time parameter; Public default is NetCorePublic-Pool; Internal default is NetCoreInternal-Pool
-            demands: ImageOverride -equals Build.Ubuntu.1604.Amd64.Open
+            demands: ImageOverride -equals Build.Ubuntu.1804.Amd64.Open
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -114,7 +114,7 @@ stages:
             name: $(PoolProvider) # This is a queue-time parameter; Public default is NetCorePublic-Pool; Internal default is NetCoreInternal-Pool
             demands: ImageOverride -equals Build.Ubuntu.1604.Amd64.Open
           ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
-            name: Hosted Ubuntu 1604
+            vmImage: ubuntu-latest
```
MattGal, line 117: Can probably update the Build.Ubuntu.1604.Amd64.Open above this too.

riarenas: Sure. the problem with staying in a fixed hosted version is this will happen again. The builds that fail with this error are not actually "failing" but are cancelled. So all our automation fails to detect it. I'd rather we break if "latest" becomes incompatible.

riarenas: Sounds good!. I'll update PRs to use the 1804 image and keep official in latest across branches.

PR #2726, opened 2021-11-13
Add ValidatingBarIds tag to Arcade Validation builds

missymessa:
Part of the work for https://github.com/dotnet/core-eng/issues/12374

The ValidatingBarIds tag in Arcade Validation denotes the BAR ID of the version of Arcade it is validating. 

This is what it looks like:
image


eng/create-baridtag.ps1
```diff
@@ -0,0 +1,25 @@
...
+. $PSScriptRoot\common\tools.ps1
+$darc = & "$PSScriptRoot\validation\get-darc.ps1"
+
+$global:azdoToken = $azdoToken
+$global:githubPAT = $githubPAT
+$global:barToken = $barToken
+
+$global:arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
+$global:arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$global:arcadeSdkPackageName
+$global:azdoRepoName = "dotnet-arcade"
```
adiaaida, line 19: This doesn't look like it's being used

adiaaida (approved):

azure-pipelines.yml
```diff
@@ -44,6 +44,28 @@ resources:
...
 
 stages:
+- stage: Create_BAR_ID_Tag
+  displayName: Create BAR ID Tag
+  jobs:
+  - template: /eng/common/templates/job/job.yml
+    parameters:
+      name: Create_BAR_ID_Tag
+      displayName: Create BAR ID Tag
+      timeoutInMinutes: 240
```
adiaaida, line 54: Does this need to be so high?

missymessa: Hah! No. Copypasta... what do we think... 10 mins sufficient?

adiaaida: I'd just take it out and let it be the azdo default (which is an hr)

eng/create-baridtag.ps1
```diff
@@ -0,0 +1,24 @@
...
+  [Parameter(Mandatory=$true)][string] $githubPAT
+)
+
+set-strictmode -version 2.0
+$ErrorActionPreference = 'Stop'
+
+. $PSScriptRoot\common\tools.ps1
+$darc = & "$PSScriptRoot\validation\get-darc.ps1"
+
+$global:azdoToken = $azdoToken
```
riarenas, line 13: Why are all the variables global?

eng/create-baridtag.ps1
```diff
@@ -0,0 +1,24 @@
...
+$global:githubPAT = $githubPAT
+$global:barToken = $barToken
+
+$global:arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
+$global:arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$global:arcadeSdkPackageName
+$jsonAsset = & $darc get-asset --name $global:arcadeSdkPackageName --version $global:arcadeSdkVersion --github-pat $global:githubPAT --azdev-pat $global:azdoToken --password $global:bartoken --output-format json | convertFrom-Json
```
riarenas, line 19: [nit] This isn't really json right? you're already deserializing it to an object.

azure-pipelines.yml
```diff
@@ -44,6 +44,27 @@ resources:
     image: mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-14.04-cross-0cd4667-20170319080304
 
 stages:
+- stage: Create_BAR_ID_Tag
```
riarenas, line 47: I think this job should be added to the post-build stages. It can either be its own stage like this, or it could be part of the arcade promotion stage. 

As it's currently written, this is breaking the pipeline's YAML because it tries to run as part of PRs. Since it can't even queue a build, the check isn't even posted to the PR: https://dev.azure.com/dnceng/public/_build/results?buildId=1469865&view=results

eng/create-baridtag.ps1
adiaaida: Agreed. I don't think you need the global: for anything you've got in this script

riarenas (changes requested):
This new functionality should only run as part of official builds, where there will be access to the secrets necessary to look up the BAR ids.

eng/create-baridtag.ps1
missymessa: Because I stole this script from another I wrote. I can unglobalize them since there's not much to this script as the one I stole it from.

riarenas: Yep, I think we can make them locals. globals make sense when you'll write lots of functions, or when you'll be using the script from another that may have conflicts.

azure-pipelines.yml
missymessa: I can pull it out of PRs, however, I want to make sure that it will add the tag no matter what, so even if I add the parameter to make sure it runs even if a previous stage breaks, if the build cancels, will it still run?

riarenas: You can set the stage to run regardless of everything by adding a `condition: always` bit in the YAML. We have an example in arcade-services: https://github.com/dotnet/arcade-services/blob/ca66009f02980f8a1c04f398d1fcfab55be7465f/eng/deploy.yaml#L251-L257
It makes sense to make it its own stage then instead of making it part of the promotion stage.

riarenas: More info on that: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml

missymessa:
Running another validation build: https://dnceng.visualstudio.com/internal/_build/results?buildId=1470334&view=results

azure-pipelines.yml
```diff
@@ -210,6 +210,28 @@ stages:
               displayName: Sign packages
 
 - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+  - stage: Create_BAR_ID_Tag
```
adiaaida, line 213: Depending on if you care when this stage runs, you could add 

```
dependsOn: []
```
and then it would run in parallel with whatever the first stage is, and tag every non-public, non-pr build.

riarenas (approved):

azure-pipelines.yml
missymessa: I think where it is works, as this was Ricardo's recommendation. But that's a good idea in case we need it :)

PR #2860, opened 2022-01-20
add UseBasicParsing to Invoke-WebRequest calls

riarenas:
Followup to https://github.com/dotnet/arcade/issues/8368. I missed that one of my test builds failed in the promotion step.

Powershell's invoke-webRequest has a dependency on Internet Explorer being installed in the machine because of course it does. The useBasicParsing parameter is how you get around that. 

Solves failures such as the one seen here: https://dev.azure.com/dnceng/internal/_build/results?buildId=1563274&view=results


riarenas:
Test build for this: https://dev.azure.com/dnceng/internal/_build/results?buildId=1563747&view=results

PR #2895, opened 2022-02-01
Add tag for Validating AzDO build

michellemcdaniel:
Using the bar id itself means that we end up having to directly reference maestro to get all of the comparison builds (we have to go current build -> maestro -> azdo build for that bar id -> compare build -> maestro -> azdo build). Maestro can be clunky and if we can avoid using it, that's better. Since we can just get the azdo url for a given asset/build, we can avoid the code that uses this information needing maestro by just saving the build url

### DIFF
--- a/eng/create-baridtag.ps1
+++ b/eng/create-baridtag.ps1
@@ -14,7 +14,9 @@ $arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
 $arcadeSdkVersion = $GlobalJson.'msbuild-sdks'.$arcadeSdkPackageName
 $assetData = & $darc get-asset --name $arcadeSdkPackageName --version $arcadeSdkVersion --github-pat $githubPAT --azdev-pat $azdoToken --password $bartoken --output-format json | convertFrom-Json
 
-## Get the BAR Build ID for the version of Arcade we are validating
+# Get the BAR Build ID for the version of Arcade we are validating
 $barBuildId = $assetData.build.id
+$azdoBuildUrl = $assetData.build.buildLink
 
 Write-Host "##vso[build.addbuildtag]ValidatingBarIds $barBuildId"
+Write-Host "##vso[build.addbuildtag]ValidatingAzDOBuild $azdoBuildUrl"
--- a/eng/create-baridtag.ps1
+++ b/eng/create-baridtag.ps1
@@ -16,7 +16,7 @@ $assetData = & $darc get-asset --name $arcadeSdkPackageName --version $arcadeSdk
 
 # Get the BAR Build ID for the version of Arcade we are validating
 $barBuildId = $assetData.build.id
-$azdoBuildUrl = $assetData.build.buildLink
+$azdoBuildId = ($assetData.build.buildLink -split "=")[-1]
 
 Write-Host "##vso[build.addbuildtag]ValidatingBarIds $barBuildId"
-Write-Host "##vso[build.addbuildtag]ValidatingAzDOBuild $azdoBuildUrl"
+Write-Host "##vso[build.addbuildtag]ValidatingAzDOBuild $azdoBuildId"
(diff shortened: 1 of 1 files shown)


eng/create-baridtag.ps1
```diff
@@ -14,7 +14,9 @@ $arcadeSdkPackageName = 'Microsoft.DotNet.Arcade.Sdk'
...
 
-## Get the BAR Build ID for the version of Arcade we are validating
+# Get the BAR Build ID for the version of Arcade we are validating
 $barBuildId = $assetData.build.id
+$azdoBuildUrl = $assetData.build.buildLink
 
-Write-Host "##vso[build.addbuildtag]ValidatingBarIds $barBuildId"
\ No newline at end of file
+Write-Host "##vso[build.addbuildtag]ValidatingBarIds $barBuildId"
+Write-Host "##vso[build.addbuildtag]ValidatingAzDOBuild $azdoBuildUrl"
```
ChadNedzlek, line 22: Are tags allowed to be this long?  This is a pretty verbose one.  Is the build ID insufficient? Is it just the "project" part that missing (the public/internal)?  Is that something that changes, or is the validation thing always internal or always public?

adiaaida: I think it will always be internal, because only internal builds can publish to bar. What I'm most concerned about is having to maintain the parsing code in multiple places (here and in dotnet-release, where we also want to turn on this functionality). Then again. If we just care about the azdo build id, the parsing amounts to split on =, take the last thing. So it's not complicated

adiaaida: I'm convinced. I'll parse here.

ChadNedzlek: Why do we need to parse anything?  The existing tag is already the build ID.  And if it's always "internal", then we should have everything we need, right?

ChadNedzlek: Unless that's some OTHER build.id?  maybe it is? Darc is fancy.

adiaaida: Yes, it is some OTHER build. Arcade validation is triggered by updates to arcade. What we care about here is the azdo build of the arcade build that triggered this arcade validation run. What we are trying to get is the commit diff between arcade now and arcade previously passing in arcade validation.

adiaaida: build.id is the BAR id (so maestro). You asked in the other PR to not use maestro, so I'm trying to get the AzDO build id.

ChadNedzlek: So "buid.url" and "build.link" are pointing to... different builds?  :-|

adiaaida: Thoughts on making darc just give me the azdo build id, instead of parsing the link?

adiaaida: so assetData.Build.Id = the BAR id for the build that triggered this build
assetData.Build.buildLink = the azdo build of the build that triggered this build

ChadNedzlek: I don't really know what darc is trying to do... like I have no idea why there are 2 builds associated in the first place, so you'd probably need to ask someone more familiar with darc.

missymessa: Given that we've been having issues with AzDO being inconsistent with the data they've been giving us recently, if we can get the value directly from darc without having to string split, I vote we go that way. 

Re: Why two builds: 

We're going to have the parent build (Arcade Validation) and then the build that it's validating (Arcade). We need to know the build ID information for those so we can do our figuring out of who committed changes to Arcade since the last time Arcade Validation passed if it starts failing. This way, we can tag those folks on the issue that Build Monitor opens and ask them to investigate the issue. Hope that helps explains why we need these values in the build.

adiaaida: PR to update darc to just give us this info: https://github.com/dotnet/arcade-services/pull/1817

adiaaida: Updated this PR to use the thing I added in the darc PR. Can't be merged until we rollout unless we decide to hardcode a version number for darc (which I don't think is necessary)

MattGal (approved):

PR #3406, opened 2022-09-30
Update latest docker tags with new schema

adiaaida:
We changed the latest tags to not include the -latest suffix, so this change updates those tags to the new schema.


missymessa (approved):
LGTM shipit

PR #3960, opened 2023-10-03
Update branch to promote builds to .NET 8 Eng

AlitzelMendez:
https://github.com/dotnet/arcade/issues/14068

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,7 +24,6 @@ trigger:
   branches:
     include:
       - main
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -25,6 +25,7 @@ trigger:
     include:
       - main
       - release/3.x
```
riarenas, line 27: I think we can remove this one?

riarenas (approved):

PR #4183, opened 2024-02-07
Specify source channel for asset validation

garath:
Resolves dotnet/dnceng#1964.

The validation checks get confused if an asset with the same name and version is published more than once. This is rare, but can happen. This change adds the source-channel flag where appropriate for calls to `darc`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -231,14 +231,11 @@ stages:
           demands: ImageOverride -equals windows.vs2022.amd64
         variables:
           - group: Publish-Build-Assets
-          - group: DotNetBot-GitHub
         steps:
           - checkout: self
             clean: true
           - powershell: eng/create-baridtag.ps1
-              -azdoToken $(dn-bot-dotnet-build-rw-code-rw)
               -barToken $(MaestroAccessToken)
-              -githubPAT $(BotAccount-dotnet-bot-repo-PAT)
             displayName: Create BAR ID Tag
             name: Create_BAR_ID_Tag
   - template: eng\common\templates\post-build\post-build.yml
--- a/eng/create-baridtag.ps1
+++ b/eng/create-baridtag.ps1
@@ -1,7 +1,5 @@
 Param(
-  [Parameter(Mandatory=$true)][string] $azdoToken,
   [Parameter(Mandatory=$true)][string] $barToken,
-  [Parameter(Mandatory=$true)][string] $githubPAT,
   [string] $sourceChannelName = '.NET Eng - Validation'
 )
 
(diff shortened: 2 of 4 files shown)


riarenas (approved):

garath:
I'm not sure how to test this. The code is all mixed in with actual asset promotion, so it's not really safe to make "simple" cuts to the pipeline. I could add a whole "dry-run" flag that does everything except the permanent changes, but that seems like more work than worth. Though I'll certainly consider reviewer pushback if folks disagree.

riarenas:
I believe disabling this job: https://github.com/dotnet/arcade-validation/blob/main/azure-pipelines.yml#L299-L321 would let you run a validation build without the build being promoted.

garath:
> I believe disabling this job

The test-publishing script in the Validate_Publishing job also makes Actual Changes in a few spots like these:

https://github.com/dotnet/arcade-validation/blob/a2dc021108f46ade3ea425a8b192ce404887a392/eng/validation/test-publishing.ps1#L69

https://github.com/dotnet/arcade-validation/blob/a2dc021108f46ade3ea425a8b192ce404887a392/eng/validation/test-publishing.ps1#L80

azure-pipelines.yml
```diff
@@ -231,14 +231,11 @@ stages:
           demands: ImageOverride -equals windows.vs2022.amd64
         variables:
           - group: Publish-Build-Assets
-          - group: DotNetBot-GitHub
         steps:
           - checkout: self
             clean: true
           - powershell: eng/create-baridtag.ps1
-              -azdoToken $(dn-bot-dotnet-build-rw-code-rw)
```
missymessa, line 239: Were these other tokens just overkill for auth?

missymessa (approved):
LGTM shipit

riarenas:
> > I believe disabling this job
> 
> The test-publishing script in the Validate_Publishing job also makes Actual Changes in a few spots like these:
> 
> https://github.com/dotnet/arcade-validation/blob/a2dc021108f46ade3ea425a8b192ce404887a392/eng/validation/test-publishing.ps1#L69
> 
> https://github.com/dotnet/arcade-validation/blob/a2dc021108f46ade3ea425a8b192ce404887a392/eng/validation/test-publishing.ps1#L80

These are fine. They just cause the arcade-validation build to get put in the general testing channel, and it pushes a branch to the internal repo that gets cleaned up when everything works.

I'm also pretty OK with you checking this in and fixing afterwards if something is off.

azure-pipelines.yml
garath: Seems to be, yeah. Which I guess makes sense for `darc get-asset` as it should just be a query to the BAR database. 

I removed it to try to make testing easier but didn't ultimately use it.

missymessa: I mean, if it's not necessary, no reason to keep them around :)

PR #4249, opened 2024-03-13
[main][1ES Template] Migrate ci pipeline

AlitzelMendez:
Migrating main branch
Pipeline migrated: https://dev.azure.com/dnceng/internal/_build?definitionId=282&_a=summary
Tested in build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2401976&view=results

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -8,7 +8,7 @@ variables:
   - ${{ if notin(variables['Build.Reason'], 'Schedule') }}:
     - name: PoolProvider
       value: NetCore1ESPool-Internal
-  - ${{ if in(variables['Build.Reason'], 'Schedule') }}:
+  - ${{ else }}:
     - name: PoolProvider
       value: NetCore1ESPool-Internal-Int
 
@@ -46,8 +46,6 @@ extends:
       name: $(PoolProvider)
       image: windows.vs2022.amd64
       os: windows
-    featureFlags:
-      autoBaseline: true
       
     stages:
     - stage: build
@@ -86,6 +84,7 @@ extends:
                 arguments: -ConfigFile $(Build.SourcesDirectory)/NuGet.config -Password $Env:Token
               env:
                 Token: $(dn-bot-dnceng-artifact-feeds-rw)
+            # Use utility script to run script command dependent on agent OS.
             - script: eng\common\cibuild.cmd
                 -configuration $(_BuildConfig)
                 -prepareMachine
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -42,277 +33,239 @@ resources:
...
-              value: $[ dependencies.setupMaestroVars.outputs['setReleaseVars.BARBuildId'] ]
-            - name: skipComponentGovernanceDetection
-              value: true
-          steps:
-            - template: /eng/common/templates/post-build/setup-maestro-vars.yml
+          - job: Linux
+            container: LinuxContainer
+            pool:
+              name: $(PoolProvider)
+              image: 1es-ubuntu-2204-pt
```
dougbu, line 101: nit
```suggestion
              image: 1es-ubuntu-2204
```

garath (approved):

azure-pipelines.yml
```diff
@@ -42,277 +33,239 @@ resources:
...
-      helixRepo: dotnet/arcade-validation
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    pool:
+      name: $(PoolProvider)
+      image: windows.vs2022.amd64
+      os: windows
+    featureFlags:
+      autoBaseline: true
```
dougbu, line 50: do we want to check auto-baselining in❔ my intention w/ dotnet-helix-machines was to baseline once and then restore TSA settings. that would prevent failing builds but keep a record of any issues found

azure-pipelines.yml
AlitzelMendez: for selecting the image I have been sticking to the guidance: https://dnceng.visualstudio.com/internal/_wiki/wikis/DNCEng%20Services%20Wiki/1214/1ES-Pipeline-Template-Migration-FAQ?anchor=i%27m-getting-started-on-migrating-my-pipeline%2C-which-pool-should-i-use%3F

is there a particular reason to use the non -pt one? should we update the guidance in that case?

dougbu: the non`-pt` images are currently identical. however the `-pt` ones are just temporary. we're not in a rush but we _will_ delete them

I'll update the guidance.

riarenas: I added some stuff to the guidance just now. Feel free to reword however you like :)

dougbu: much appreciated. I'm going to reword it to _exclude_ the still-incompatible images because that list is shorter 😄 

/fyi you left out the other `windows.vs*` images. that fix (dotnet/dnceng#2080) also rolled out today

riarenas: Ah for some reason I thought that included all the windows images already. Thanks for correcting

dougbu (approved):
thanks for addressing my nits

PR #4271, opened 2024-03-19
Update Debian version in `main`

richlander:
I notice Windows 10. Should that be updated to Windows 11?


riarenas (approved):

riarenas:
> I notice Windows 10. Should that be updated to Windows 11?

Yep I think it's a good idea to update: https://github.com/dotnet/arcade/issues/14622
This repo is a little special in that we don't really care what we run, as we're validating that changes to the Helix SDK don't break scenarios, but that also means there's also no good reason to keep sending to 10.

PR #4575, opened 2024-09-24
Enable signing on Linux

ellahathaway:
Related to https://github.com/dotnet/arcade/issues/14431

Validates the changes in https://github.com/dotnet/arcade/pull/15104 for Linux (Mac is currently blocked - see https://github.com/dotnet/arcade/issues/14431#issuecomment-2387129090)

[Test pipeline run](https://dev.azure.com/dnceng/internal/_build/results?buildId=2552857&view=results) (internal Microsoft link)

### DIFF
--- a/src/Validation/src/Validation.csproj
+++ b/src/Validation/src/Validation.csproj
@@ -8,11 +8,7 @@
 
   <!-- Set package ID for official signed builds. Needed to publish both Windows and Linux signed packages to build asset registry. -->
   <PropertyGroup Condition="'$(DotNetSignType)' == 'real'">
-    <BuildRid Condition="'$(BuildRid)' == '' and '$(MSBuildRuntimeType)' == 'core'">$([System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier)</BuildRid>
-    <TargetArchitecture Condition="'$(TargetArchitecture)' == ''">$([System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant)</TargetArchitecture>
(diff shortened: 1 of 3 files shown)


src/Validation/src/Validation.csproj
```diff
@@ -4,7 +4,15 @@
     <OutputType>Exe</OutputType>
     <TargetFrameworks>$(NetCurrent)</TargetFrameworks>
     <IsPackable>true</IsPackable>
-    <SignAssembly>true</SignAssembly>
+  </PropertyGroup>
+
+  <!-- Set package ID for official signed builds. Needed to publish both Windows and Linux signed packages to build asset registry. -->
+  <PropertyGroup Condition="'$(DotNetSignType)' == 'real'">
+    <BuildRid Condition="'$(BuildRid)' == '' and '$(MSBuildRuntimeType)' == 'core'">$([System.Runtime.InteropServices.RuntimeInformation]::RuntimeIdentifier)</BuildRid>
```
mmitche, line 11: Why is it necessary to set the Target properties here? As opposed to just the package id.

ellahathaway: I suppose it's not necessary, this is similar to how we do it in the VMR: https://github.com/dotnet/sdk/blob/619b55dd198bbea2d5021ad0148f82097c379e21/src/SourceBuild/content/Directory.Build.props#L76.

I'll adjust this section to just be <PackageId>Validation.$(OS)</PackageId>

mmitche: The VMR is a bit different here because a lot of those parameters do matter for the artifacts produced. For this package, it shouldn't matter.

mmitche (approved):

PR #4855, opened 2025-03-12
Move checkout to presteps

ellahathaway:
Related to https://github.com/dotnet/source-build/issues/4793.

MicroBuild install has to run after the repo checkout since it installs to the sources directory by default (see https://github.com/dotnet/arcade/pull/15626/commits/0da7c5f1373544b37fa9ca59d3d79bde9adb7b55). For context, it needs to install to the sources directory because the infra requires .NET 8 during signing.

[Internal test build](https://dev.azure.com/dnceng/internal/_build/results?buildId=2661574&view=results)

### DIFF
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -46,9 +46,10 @@ stages:
             Build_Debug:
               _BuildConfig: Debug
               _SignType: test
-        steps:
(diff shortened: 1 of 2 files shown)


missymessa (approved):
LGTM shipit

PR #4888, opened 2025-04-03
Removing EOL Software

ilyas1974:
Replacing Ubuntu 18.04 with Ubuntu 22.04, Mariner 2 with Azure Linux 3 and Debian 10 with Debian 12


ilyas1974:
wrong branch

PR #5195, opened 2025-10-02
Create a pipeline to test preview microbuild plugin against arcade va…

mmitche:
…lidation

- Core-templates to be updated in separate arcade PR. Here for illustration
- Factor out the main build stage into a separate YAML so that we don't duplicate that in the new preview YAML
- Add preview pipeline YAML, with non-interesting steps removed (Helix, publishing, etc.)


eng/common/core-templates/steps/install-microbuild-impl.yml
```diff
@@ -0,0 +1,34 @@
+parameters:
```
ellahathaway, line 1: Will touching `eng/common` cause issues when darc tries to update this repo with the changes from https://github.com/dotnet/arcade/pull/16185?

mmitche: I'm going to wait till that comes in, then merge in the latest here. Should then be no-diff.

